PR #1, opened 2024-01-30
FIX: Allow Path or string to download routine

greglucas:
# Change Summary

## Overview

Passing a pathlib object before caused errors on the `"/" in file_path` check due to it not being a string. We should only go down that path if we are passed strings, so add that case and add an explicit test for passing Path objects.

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -14,12 +14,12 @@ import imap_data_access
 logger = logging.getLogger(__name__)
 
 
-def download(file_path: str) -> Path:
+def download(file_path: Path | str) -> Path:
     """Download a file from the data archive.
 
     Parameters
     ----------
-    file_path : str
+    file_path : pathlib.Path or str
         Name of the file to download, optionally including the directory path
 
     Returns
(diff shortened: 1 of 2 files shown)


imap_data_access/io.py
```diff
@@ -28,7 +28,7 @@ def download(file_path: str) -> Path:
         Path to the downloaded file
     """
     destination = imap_data_access.config["DATA_DIR"]
-    if "/" not in file_path:
+    if isinstance(file_path, str) and "/" not in file_path:
```
tech3371, line 31: will this only check if path is string or does it convert Path object to str for the check?

greglucas: It see if it is a string, then if it is a string checks if "/" is in that string, and if it isn't then it goes down this if-block. (we have only been given a filename, not the full path).

We could tighten this up if we wanted and remove this altogether and require the full path even for the download routines?

tech3371: I feel like it should something like:
```
if file_path is Path object, then convert to str. Then check for '/' in file_path.
else if file_path is str, then '/' in file_path
```
The issue was that download would fail in input is path object because we don't convert that to str and as a result `if "/" not in file_path:` check fails.

tech3371: I see. I just saw your reply. I think that makes sense.

tech3371 (approved):

PR #2, opened 2024-01-30
DOC: Update README with some more information

greglucas:
- Move command line access up to the top for quick reference.
- Put the configuration options at the bottom.
- Add a section on networking issues


README.md
```diff
@@ -2,17 +2,30 @@
...
+### Query / Search for data
 
-A user's root data location can be specified as an environment
-variable ``IMAP_DATA_DIR`` or through a configuration dictionary
-within the package itself ``imap_data_access.config["DATA_DIR"]``.
+```bash
+# Find all files from the SWE instrument
+imap-data-access query --instrument swe
+# Find all files during the year 2025
+imap-data-access query --start_date 20250101 --end_date 20251231
```
bourque, line 13: We might want to use parameters here that actually return results, otherwise users might get discouraged or think something is wrong if they get an empty result?

README.md
```diff
@@ -2,17 +2,30 @@
 
 This is a minimal Python package that allows users to download, query, and upload data from the IMAP Science Data Center (SDC).
```
bourque, line 3:
```suggestion
This lightweight Python package allows users to download, query, and upload data from the IMAP Science Data Center (SDC).
```

For some reason I feel that using the word 'minimal' here comes off as diminishing of the capability of this package. Users might think 'this is not going to be good enough to handle my complex query'.

README.md
```diff
@@ -31,3 +44,51 @@ imap_data_access.download("imap/mag/l0/2024/01/imap_mag_l0_raw_202040101_2024010
 # Upload a calibration file that exists locally
 imap_data_access.upload("imap/mag/calibration/test_calibration.txt")
 ```
+
+## Configuration
```
bourque, line 48: I would actually prefer this section to be at the beginning, otherwise users might try to upload a file without first realizing that they need to have a directory structure/`DATA_DIR` env variable configured. But this could be a case of my brain just reproaching the situation different from yours -- we are both right and both wrong 😅

README.md
```diff
@@ -31,3 +44,51 @@ imap_data_access.download("imap/mag/l0/2024/01/imap_mag_l0_raw_202040101_2024010
...
+
+### Data directory
```
bourque, line 50: It might be helpful to include an example of this structure, e.g.:

```
The directory structure is as follows:

imap/
  <instrument>/
    <data_level>/
      <year>/
        <month>/
          <filename>

for example:

imap/
  swe/
    l0/
      2024/
        01/
          imap_swe_l0_sci_20240105_20240105_v00-00.pkts
```

README.md
```diff
@@ -31,3 +44,51 @@ imap_data_access.download("imap/mag/l0/2024/01/imap_mag_l0_raw_202040101_2024010
...
+
+### Network issues
+
+If you encounter SSL errors similar to the following:
+
+```text
+urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>
+```
+
+That generally means the Python you're using is not finding your system's root
```
bourque, line 79:
```suggestion
That generally means the Python environment you're using is not finding your system's root
```

README.md
```diff
@@ -31,3 +44,51 @@ imap_data_access.download("imap/mag/l0/2024/01/imap_mag_l0_raw_202040101_2024010
...
+    ```bash
+    pip install --upgrade certifi
+    ```
+
+2. Install system certificates
+    Depending on the Python version you installed the program with the command will look something like this:
+
+    ```bash
+    /Applications/Python\ 3.10/Install\ Certificates.command
+    ```
```
bourque, line 94: Two other troubleshooting cases we could add:

1. `urllib.error.HTTPError: HTTP Error 502: Bad Gateway` - I came across this when I tried a download, I think because the AWS `dev` stacks are currently being re-deployed? But it might be good to add a note that if users see this, to contact the SDC.
2. `FileNotFoundError` could indicate that the user does not have the data directory set up properly.

README.md
```diff
@@ -31,3 +44,51 @@ imap_data_access.download("imap/mag/l0/2024/01/imap_mag_l0_raw_202040101_2024010
...
+within the package itself ``imap_data_access.config["DATA_DIR"]``.
+
+### Data Access URL
+
+To change the default URL that the package accesses, you can set
+the environment variable ``IMAP_DATA_ACCESS_URL`` or within the
+package ``imap_data_access.config["DATA_ACCESS_URL"]``. The default
+is the development server ``https://api.dev.imap-mission.com``.
+
+## Troubleshooting
```
bourque, line 69: Great idea to add a troubleshooting section!

bourque (commented):
This is great, and I can see how useful this repo is going to be already! Nice work. I have some comments/suggestions.

README.md
greglucas: I had this first before, then decided to change it! :) So I can see both ways the same as you. I was thinking people often just want to start using a tool, which is why I put the cli thing up front. I'm happy to move this back up though if you think that is better that way. I don't have a major preference.

README.md
greglucas: Good suggestion, I updated to 2024

README.md
greglucas: I like this suggestion, thanks!

README.md
greglucas: Added both of these, all still under Networking. Not sure if we want to move them around at all or put differnet sub-headings.

README.md
bourque: Yeah I can't say I have a major preference either, so I will leave it up to you. This is when it would be nice to have a product owner who can make these end-user decisions for us 😂

README.md
```diff
@@ -1,18 +1,31 @@
...
+```bash
+# Level 0 SWE file on 2024/01/05
+imap-data-access download imap/swe/l0/2024/01/imap_swe_l0_sci_20240105_20240105_v00-00.pkts
+```
+
+### Upload a file
+
+```bash
+# Upload an l1a file after decoding the l0 CCSDS ".pkts" file
+imap-data-access upload imap/swe/l0/2024/01/imap_swe_l1a_sci_20240105_20240105_v00-00.cdf
```
bourque, line 27:
```suggestion
imap-data-access upload imap/swe/l1a/2024/01/imap_swe_l1a_sci_20240105_20240105_v00-00.cdf
```

bourque (approved):
Looks good!

README.md
greglucas: I left it as-is. Happy to add more content / rearrange later in follow-up PRs too.

PR #6, opened 2024-01-31
MNT: Cleanup cli prints

greglucas:
- extra debug print hanging around
- There is no output from the upload function, so print the original file

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -58,7 +58,6 @@ def _query_parser(args: argparse.Namespace):
     }
     output = imap_data_access.query(**query_params)
     num_files = len(output)
-    print(type(output))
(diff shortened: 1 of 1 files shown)


imap_data_access/cli.py
```diff
@@ -58,7 +58,6 @@ def _query_parser(args: argparse.Namespace):
     }
     output = imap_data_access.query(**query_params)
     num_files = len(output)
-    print(type(output))
     print(f"Found [{num_files}] files matching {query_params}:")
     print(output)
```
greglucas, line 62: Is the first line with what we found even useful, or should we just return the results so it can be piped somewhere else?

bourque (approved):
LGTM!

PR #7, opened 2024-01-31
FIX: Allow relative paths and strings to be input to upload

greglucas:
This allows relative paths and strings to be uploaded. Before the routines required the absolute path even if you were in the data directory.

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -156,15 +156,16 @@ def query(
     return items
 
 
-def upload(file_path: Path) -> None:
+def upload(file_path: Union[Path, str]) -> None:
     """Upload a file to the data archive.
 
     Parameters
     ----------
-    file_path : pathlib.Path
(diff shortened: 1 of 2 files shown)


imap_data_access/io.py
```diff
@@ -175,11 +176,14 @@ def upload(file_path: Path) -> None:
...
-    upload_name = str(file_path.relative_to(imap_data_access.config["DATA_DIR"]))
+    # Must be posix style for the URL
+    upload_name = str(
+        file_path.relative_to(imap_data_access.config["DATA_DIR"]).as_posix()
+    )
 
     url = f"{imap_data_access.config['DATA_ACCESS_URL']}"
     # The upload name needs to be given as a path parameter
-    url += f"/upload/science/{upload_name}"
+    url += f"/upload/{upload_name}"
```
greglucas, line 186: Realized that we had the `science/` leftover here still, so I removed it now.

PR #8, opened 2024-01-31
MNT: Updates to command line arguments

greglucas:
- change hyphen to underscores in command line arguments
- add a table and json output options
- update the README with some explicit examples

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -16,6 +16,7 @@ Use
 """
 import argparse
 import logging
+import os
 from pathlib import Path
 
 import imap_data_access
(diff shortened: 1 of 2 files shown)


imap_data_access/cli.py
```diff
@@ -33,6 +34,50 @@ def _download_parser(args: argparse.Namespace):
     print(f"Downloaded the file to: {output_path}")
 
 
+def _print_query_results_table(query_results):
```
bourque, line 37: What you have here works just fine, and looks nice! I wonder if you could achieve a similar result with simpler/fewer lines of code by using a dataframe? Something along the lines of this: https://chat.openai.com/share/61f0f4e5-2ea9-482e-a424-86be1896e6d4

imap_data_access/cli.py
```diff
@@ -182,6 +230,14 @@ def main():
     query_parser.add_argument(
         "--extension", type=str, required=False, help="File extension (cdf, pkts, etc.)"
     )
+    query_parser.add_argument(
+        "--output-format",
+        type=str,
+        required=False,
+        help="How to format the output, default is 'table'",
+        choices=["table", "json"],
```
bourque, line 238: Somehow I have never seen this `choices` parameter before. That's neat. Does this automatically raise an error if a user supplies something that is not in this list?

bourque (approved):
Nice work, love the table!

imap_data_access/cli.py
greglucas: I agree, this could definitely be done easier with external libraries 😂 . But, I was really striving for this package to have 0 dependencies and just be able to be installed/used as a script with no fuss.

bourque: Yeah, I can see the benefit to that. This seems to work fine, so happy to go with it.

imap_data_access/cli.py
greglucas: Yep, and the help even gives the list of options.
```
imap-data-access query --instrument abc
usage: imap-data-access query [-h]
                              [--instrument {codice,glows,hi,hit,idex,lo,mag,swapi,swe,ultra}]
                              [--data-level DATA_LEVEL]
                              [--descriptor DESCRIPTOR]
                              [--start-date START_DATE] [--end-date END_DATE]
                              [--version VERSION] [--extension EXTENSION]
                              [--output-format {table,json}]
imap-data-access query: error: argument --instrument: invalid choice: 'abc' (choose from 'codice', 'glows', 'hi', 'hit', 'idex', 'lo', 'mag', 'swapi', 'swe', 'ultra')
```

imap_data_access/cli.py
greglucas: We can have a summer student make this nicer later :)

PR #9, opened 2024-01-31
FIX: Catch errors when using the CLI

greglucas:
This avoids the stack traces and only prints the messages of what hapened with the server errors instead.

Are these print statements what you'd expect to see as a user?
```bash
$ imap-data-access download a/b/c.txt
HTTP Error: 502 - Bad Gateway
Server Message: {"message": "Internal server error"}
```


bourque (approved):

sdhoyt:
Sorry, just seeing this now. That's weird that it was bad gateway,
download api should return this if there's not file found:
```python
http_response(status_code=404, body="File not found in S3.")
```

PR #11, opened 2024-02-03
Updates to README examples, IMAP_DATA_DIR clarification

bourque:
# Change Summary

## Overview
This PR makes minor updates to the README:
- Updated upload example to avoid use of calibration files, which are currently not supported
- Added sentence about what happens if user does not set the `IMAP_DATA_DIR` env variable


greglucas (approved):
Thanks!

PR #12, opened 2024-02-05
Adding filename utility filename

maxinelasp:
# Change Summary
This is a proposal for this data-access repo - that we should move various filename related utilities out of SDS data manager/imap_processing and into this repo. Then it could be imported into both projects without duplicating code or needing to import large sds-data-manager into imap-processing. This could also be used to make additional command line utilities, such as `--validate filename`.

## Overview

For now, just added a split method to break a file name into multiple components. In the future, I think moving [path_helper.py](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/lambda_code/SDSCode/path_helper.py) into this project would make sense.

## New Files

- file_naming.py
   - The name of this file sucks, but I didn't want to name it `file_utils.py`
   - copied extract_filename method from [batch_starter.py](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/84e74b576b8f4e622d3d5619a62e9a294f8b1f71/sds_data_manager/lambda_code/SDSCode/batch_starter.py#L325)

### DIFF
--- a/imap_data_access/file_naming.py
+++ b/imap_data_access/file_naming.py
@@ -18,9 +18,9 @@ def extract_filename_components(filename: str):
     """
     pattern = (
         r"^imap_"
-        r"(?P<instrument>[^_]*)_"
-        r"(?P<datalevel>[^_]*)_"
-        r"(?P<descriptor>[^_]*)_"
+        r"(?P<instrument>[^_]+)_"
+        r"(?P<datalevel>[^_]+)_"
+        r"(?P<descriptor>[^_]*)_?"  # optional
         r"(?P<startdate>\d{8})_"
         r"(?P<enddate>\d{8})_"
         r"(?P<version>v\d{2}-\d{2})"
--- a/imap_data_access/file_naming.py
+++ b/imap_data_access/file_naming.py
@@ -1,0 +1,33 @@
+"""Methods for managing and validating filenames."""
+import re
+
+def extract_filename_components(filename: str):
+    """
+    Extracts all components from filename.
+
+    Parameters
+    ----------
+    filename : str
+        Path of dependency data.
+
+    Returns
+    -------
+    components : dict
+        Dictionary containing components.
+
+    """
+    pattern = (
+        r"^imap_"
+        r"(?P<instrument>[^_]*)_"
+        r"(?P<datalevel>[^_]*)_"
+        r"(?P<descriptor>[^_]*)_"
+        r"(?P<startdate>\d{8})_"
+        r"(?P<enddate>\d{8})_"
+        r"(?P<version>v\d{2}-\d{2})"
+        r"\.(cdf|pkts)$"
+    )
+    match = re.match(pattern, filename)
+    if match is None:
+        return
+    components = match.groupdict()
+    return components


imap_data_access/file_naming.py
```diff
@@ -0,0 +1,33 @@
...
+import re
+
+def extract_filename_components(filename: str):
+    """
+    Extracts all components from filename.
+
+    Parameters
+    ----------
+    filename : str
+        Path of dependency data.
```
greglucas, line 11: Do we want to accept paths here too? Might be able to handle them easily enough with: `filename = Path(filename).name`

```suggestion
    filename : pathlib.Path or str
        Name of the file to extract components from
```

imap_data_access/file_naming.py
```diff
@@ -0,0 +1,33 @@
+"""Methods for managing and validating filenames."""
+import re
+
+def extract_filename_components(filename: str):
+    """
+    Extracts all components from filename.
```
greglucas, line 6: Would be nice to list the components explicitly in the docstring here.

imap_data_access/file_naming.py
```diff
@@ -0,0 +1,33 @@
...
+        r"(?P<datalevel>[^_]+)_"
+        r"(?P<descriptor>[^_]*)_?"  # optional
+        r"(?P<startdate>\d{8})_"
+        r"(?P<enddate>\d{8})_"
+        r"(?P<version>v\d{2}-\d{2})"
+        r"\.(cdf|pkts)$"
+    )
+    match = re.match(pattern, filename)
+    if match is None:
+        return
```
greglucas, line 31: Do we want to raise an error here instead indicating its a bad filename? I think path_helper.py might have more checks and put some helpful messages in if it doesn't match the structure, do we want to bring those over as well?

greglucas (commented):
I think this is a great idea :+1: Needs some tests too.

For naming possibly `file_validation.py` ? or something indicating these are validation helpers.

maxinelasp:
@greglucas Agreed with your notes, I wanted to open the discussion and see what people thought before adding more stuff here.

maxinelasp:
@tech3371 , you wrote most/all of pathhelper, what do you think about moving it here and making it a common utility across sds_data_manager and imap-processing?

bourque:
Along the same lines of not repeating code across `sds-data-manager` & `imap_processing`, would this repo be a good place to store stuff like valid instrument names and data levels (i.e. what is listed [here](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/a29978f3a31d5f1ff10f256245e5eb4ab20a8b37/imap_processing/__init__.py#L34)) ?

maxinelasp:
> Along the same lines of not repeating code across sds-data-manager & imap_processing, would this repo be a good place to store stuff like valid instrument names and data levels (i.e. what is listed [here](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/a29978f3a31d5f1ff10f256245e5eb4ab20a8b37/imap_processing/__init__.py#L34)) ?

I think this makes a lot of sense, especially as it's part of checking a valid file name. It's not super related to data access beyond that... but also it's a minor thing that would be really easy to add here and just reference everywhere, so ... I like that idea.

tech3371:
I haven't given much though but yeah, I like the idea of keeping things that's commonly used in this library.

maxinelasp:
Since I'm hearing some positive info, I will add some of the filename utilities into this repo and open a more comprehensive code review. Then, we can remove some code from sds-data-manager and imap_processing

tech3371:
@maxinelasp I was going to ask if you could help me improve path_helper.py class. If you could improve it during this movement, I appreciate.

PR #14, opened 2024-02-07
Adding universally useful file path and file name management to imap-data-access

maxinelasp:
# Change Summary
Moving some of the file management stuff from sds-data-manager to this repo for wider use. I'd appreciate a close review from @bourque, @greglucas, and @tech3371 to make sure this lines up with what they are expecting

## Overview
Added new file_validation.py file, with a shared ScienceFilepathManager class (copied from Tenzin's work on sds-manager, I need to add her and Laura as coauthors actually), which I refactored slightly for more wide adoption. Also added some useful shared methods in that file and a list of valid instruments and levels. 

## New Dependencies
I moved the project to poetry (don't leave me unattended) to match the other projects. No new dependencies there though

## New Files

## Updated Files

- imap_data_access/io.py
   - updated to move some logic out of the function and use the new shared functionality. I would appreciate a closer look at these changes in case I messed up the way the expected path works.
   
## Testing

Updated one test to add `imap/` as part of the expected path... I think this is correct, but would appreciate a double-check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,3 +148,26 @@ continue to encounter this, reach out to the IMAP SDC at
 This could mean that the local data directory is not set
 up with the same paths as the SDC. See the [data directory](#data-directory)
 section for an example of how to set this up.
+
+## File Validation
+
+This package validates filenames and paths to check they follow our standards, as defined by the filename conventions. There is also a class available for
+use by other packages to create filepaths and filenames that follow the IMAP SDC conventions.
+
+To use this class, use `imap_data_access.ScienceFilepath`.
+
+Usage:
+
+```python
+
+science_file = imap_data_access.ScienceFilePath("imap_swe_l0_sci_20240105_20240105_v00-05.pkts")
+
+# Filepath = /imap/swe/l0/2024/01/imap_swe_l0_sci_20240105_20240105_v00-05.pkts
+filepath = science_file.construct_file_path()
+
+# Use a data directory
+science_file = imap_data_access.ScienceFilePath("imap_swe_l0_sci_20240105_20240105_v00-05.pkts", data_dir="/data/personal")
+
+# Filepath = /data/personal/imap/swe/l0/2024/01/imap_swe_l0_sci_20240105_20240105_v00-05.pkts
+filepath = science_file.construct_file_path()
+```
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -8,8 +8,9 @@ import os
 from pathlib import Path
 
 from imap_data_access.io import download, query, upload
+from imap_data_access.file_validation import construct_upload_path
 
-__all__ = ["query", "download", "upload"]
+__all__ = ["query", "download", "upload", "construct_upload_path"]
 __version__ = "0.2.0"
 
 
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -7,10 +7,10 @@ provides a convenient way to query the IMAP data archive and download data files
 import os
 from pathlib import Path
 
-from imap_data_access.file_validation import ScienceFilepath
+from imap_data_access.file_validation import ScienceFilePath
 from imap_data_access.io import download, query, upload
 
-__all__ = ["query", "download", "upload", "ScienceFilepath"]
+__all__ = ["query", "download", "upload", "ScienceFilePath"]
 __version__ = "0.2.0"
 
 
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -7,10 +7,10 @@ provides a convenient way to query the IMAP data archive and download data files
 import os
 from pathlib import Path
 
-from imap_data_access.file_validation import construct_path_from_filename
+from imap_data_access.file_validation import ScienceFilepath
 from imap_data_access.io import download, query, upload
 
-__all__ = ["query", "download", "upload", "construct_path_from_filename"]
+__all__ = ["query", "download", "upload", "ScienceFilepath"]
 __version__ = "0.2.0"
 
 
@@ -44,7 +44,22 @@ VALID_INSTRUMENTS = {
     "ultra-90",
 }
 
-VALID_DATALEVELS = {"l0", "l1", "l1a", "l1b", "l1c", "l1d", "l2"}
+VALID_DATALEVELS = {
+    "l0",
+    "l1",
+    "l1a",
+    "l1b",
+    "l1c",
+    "l1ca",
+    "l1cb" "l1d",
+    "l2",
+    "l2pre",
+    "l3",
+    "l3a",
+    "l3b",
+    "l3c",
+    "l3d",
+}
 
 VALID_FILE_EXTENSION = {"pkts", "cdf"}
 
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -7,10 +7,10 @@ provides a convenient way to query the IMAP data archive and download data files
 import os
 from pathlib import Path
 
+from imap_data_access.file_validation import construct_path_from_filename
 from imap_data_access.io import download, query, upload
-from imap_data_access.file_validation import construct_upload_path
 
-__all__ = ["query", "download", "upload", "construct_upload_path"]
+__all__ = ["query", "download", "upload", "construct_path_from_filename"]
 __version__ = "0.2.0"
 
 
@@ -27,3 +27,28 @@ DATA_DIR : This is where the file data is stored and organized by instrument and
     "but this can be set on the command line using the --data-dir option, or through
     the environment variable IMAP_DATA_DIR.
 """
+
+
+VALID_INSTRUMENTS = {
+    "codice",
+    "glows",
+    "hit",
+    "hi-45",
+    "hi-90",
+    "idex",
+    "lo",
+    "mag",
+    "swapi",
+    "swe",
+    "ultra-45",
+    "ultra-90",
+}
+
+VALID_DATALEVELS = {"l0", "l1", "l1a", "l1b", "l1c", "l1d", "l2"}
+
+VALID_FILE_EXTENSION = {"pkts", "cdf"}
+
+FILENAME_CONVENTION = (
+    "<mission>_<instrument>_<datalevel>_<descriptor>_"
+    "<startdate>_<enddate>_<version>.<extension>"
+)
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -8,93 +8,21 @@ from pathlib import Path
 import imap_data_access
 
 
-def extract_filename_components(filename: str | Path):
-    """
-    Extracts all components from filename. Does not validate instrument or level.
+class ScienceFilepath:
+    class InvalidScienceFileError(Exception):
+        """Indicates a bad file type"""
 
-    Will return a dictionary with the following keys:
-    { instrument, datalevel, descriptor, startdate, enddate, version, extension }
+        pass
 
-    If a match is not found, a ValueError will be raised.
+    def __init__(self, filename: str | Path, data_dir: Path | None = None):
+        """Class to store filepath and file management methods for science files.
 
-    Parameters
-    ----------
-    filename : Path or str
-        Path of dependency data.
-
-    Returns
-    -------
-    components : dict
-        Dictionary containing components.
-
-    """
-    pattern = (
-        r"^(?P<mission>imap)_"
-        r"(?P<instrument>[^_]+)_"
-        r"(?P<datalevel>[^_]+)_"
-        r"(?P<descriptor>[^_]+)_"
-        r"(?P<startdate>\d{8})_"
-        r"(?P<enddate>\d{8})_"
-        r"(?P<version>v\d{2}-\d{2})"
-        r"\.(?P<extension>cdf|pkts)$"
-    )
-    if isinstance(filename, Path):
-        filename = filename.name
-
-    match = re.match(pattern, filename)
-    if match is None:
-        raise ValueError(
-            f"Filename {filename} does not match expected pattern: "
-            f"{imap_data_access.FILENAME_CONVENTION}"
-        )
-
-    components = match.groupdict()
-    return components
-
-
-def construct_path_from_filename(filename: str | Path) -> Path:
-    """
-    Given the filename, construct the expected upload path.
-
-    Example path: imap/mag/l1a/2021/01/imap_mag_l1a_burst_20210101_20210102_v01-01.cdf
-
-    Parameters
-    ----------
-    filename: str | Path
-        File name or path to create the path from. Should match IMAP file convention:
-        <mission>_<instrument>_<datalevel>_<descriptor>_<startdate>_<enddate>_<version>.<extension>
-
-    Returns
-    -------
-    Path
-        File path generated from the filename.
-
-    """
-
-    components = extract_filename_components(filename)
-
-    return (
-        Path(
-            f"{components['mission']}/{components['instrument']}/"
-            f"{components['datalevel']}/{components['startdate'][:4]}/"
-            f"{components['startdate'][4:6]}"
-        )
-        / filename
-    )
-
-
-class InvalidScienceFileError(Exception):
-    """Indicates a bad file type"""
-
-    pass
-
-
-class ScienceFilepathManager:
-    def __init__(self, filename: str | Path):
-        """Class to store file pattern
+        If you have an instance of this class, you can be confident you have a valid
+        science file and generate paths in the correct format.
 
         Current filename convention:
-        <mission>_<instrument>_<datalevel>_<descriptor>_<startdate>_<enddate>_<version>.<extension>
+        <mission>_<instrument>_<datalevel>_<descriptor>_<startdate>_<enddate>
+        _<version>.<extension>
 
         NOTE: There are no optional parameters anymore. All parameters are required.
         <mission>: imap
@@ -113,14 +41,16 @@ class ScienceFilepathManager:
         ----------
         filename : str | Path
             Science data filename or file path.
+        data_dir : Path, optional
+            Optional higher directory level for the data, by default None
         """
-        # TODO: Accomodate path or filename
-        self.filename = filename
+        self.filename = Path(filename)
+        self.data_dir = data_dir
 
         try:
-            split_filename = extract_filename_components(self.filename)
+            split_filename = self.extract_filename_components(self.filename)
         except ValueError as err:
-            raise InvalidScienceFileError(
+            raise self.InvalidScienceFileError(
                 f"Invalid filename. Expected file to match format: "
                 f"{imap_data_access.FILENAME_CONVENTION}"
             ) from err
@@ -136,9 +66,9 @@ class ScienceFilepathManager:
 
         self.error_message = self.validate_filename()
         if self.error_message:
-            raise InvalidScienceFileError(f"{self.error_message}")
+            raise self.InvalidScienceFileError(f"{self.error_message}")
 
-    def validate_filename(self):
+    def validate_filename(self) -> str:
         """Validate the filename and populate the error message for wrong attributes.
 
         The error message will be an empty string if the filename is valid. Otherwise,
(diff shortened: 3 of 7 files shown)


imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Will return a dictionary with the following keys:
+    { instrument, datalevel, descriptor, startdate, enddate, version, extension }
+
+    If a match is not found, a ValueError will be raised.
+
+    Parameters
+    ----------
+    filename : Path or str
+        Path of dependency data.
```
greglucas, line 46:
```suggestion
        Path to the file.
```

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Returns
+    -------
+    components : dict
+        Dictionary containing components.
```
greglucas, line 51:
```suggestion
    dict
        Components of the filename
```

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+        r"(?P<enddate>\d{8})_"
+        r"(?P<version>v\d{2}-\d{2})"
+        r"\.(cdf|pkts)$"
+    )
+    if isinstance(filename, Path):
+        filename = filename.name
+
+    match = re.match(pattern, filename)
+    if match is None:
+        raise ValueError(f"Filename {filename} does not match expected pattern")
```
greglucas, line 69:
```suggestion
        raise ValueError(f"Filename {filename} does not match the expected pattern: {FILENAME_CONVENTION}")
```
or give an example?

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+
+    components = match.groupdict()
+    components["extension"] = match.group(7)
+    components["mission"] = "imap"
```
greglucas, line 73: I think you can add these labels into the regex as well to avoid the special cases.

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+    return components
+
+
+def construct_upload_path(filename: str | Path) -> Path:
```
greglucas, line 77: I think this might be used in places other than just upload, so something more generic might be good in the name, `construct_file_path_from_name` (a little long, so not super ideal, but just an idea)

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+        Path(
+            f"{components['mission']}/{components['instrument']}/"
+            f"{components['datalevel']}/{components['startdate'][:4]}/"
+            f"{components['startdate'][4:6]}"
+        )
+        / filename
+    )
+
+
+class InvalidScienceFileError(Exception):
```
greglucas, line 108: I'm a bit confused from here on down. I think these are two different ways of approaching the same problem, so are we choosing one or the other, or using both for some reason?

imap_data_access/io.py
```diff
@@ -64,14 +64,12 @@ def download(file_path: Union[Path, str]) -> Path:
...
-        year = startdate[:4]
-        month = startdate[4:6]
-        destination = destination / instrument / datalevel / year / month
-    destination /= file_path
+
+        construct = imap_data_access.construct_upload_path(file_path)
+
+        destination = destination / construct
+    else:
+        destination /= file_path
```
greglucas, line 72: There was discussion about removing this special path for science files as well and forcing users to use the full path structure. Thoughts on removing this path entirely?

pyproject.toml
greglucas, line 1: Would you be able to pull the poetry stuff out into a separate PR?

I am torn on it here. I realize it would be nice for consistency across all our repos, but it adds a lot of bloat to a simple package with no dependencies IMO. One of my hangups is that they don't follow the recent PEPs and use the standard-defined `[project]` sections which makes it hard to move back and forth between build-backends (as seen here there is a lot of churn in this file). I'm personally not sold at all on Poetry for distributing packages (users can still use poetry themselves no problem it just wouldn't be the build backend), but I understand we are using it in the other projects so it might be good for consistency. All that to say, I think it'd be good to put in a separate PR to separate discussions.

pyproject.toml
```diff
@@ -1,52 +1,36 @@
...
 testpaths = [
   "tests",
 ]
 addopts = "-ra"
-
-[tool.ruff]
-target-version = "py39"
-select = ["B", "E", "F", "I", "N", "W", "PL", "PT", "UP", "RUF"]
-ignore = ["PLR2004"]
+markers = [
```
greglucas, line 29: We should leave the ruff section and remove the unused pieces that were added here.

tests/test_io.py
```diff
@@ -48,7 +48,7 @@ def test_request_errors(mock_urlopen):
         # Directory structure inferred
         (
             "imap_test_l1_test-description_20100101_20100102_v00-00.cdf",
-            "test/l1/2010/01/imap_test_l1_test-description_20100101_20100102_v00-00.cdf",
+            "imap/test/l1/2010/01/imap_test_l1_test-description_20100101_20100102_v00-00.cdf",
```
greglucas, line 51: Looks like this caught an error! Do we want to put a test for the previous bad download path here as well? It might already be covered in the other file_validation tests though, so probably not a huge reason to add it here.

greglucas (commented):
I'm a bit confused with the mixing of the functions for checking versus using the class for checking here. My reading of it is that we'd only need one or the other, but maybe I'm missing something... I started reviewing the function form and then hit the class and didn't review that because I figured it'd be good to get clarification on the intended usages first.

pyproject.toml
maxinelasp: Sure, I can move it. I'm not sure what you mean by bloat though - just that developers need to `pip install` poetry itself for dependencies...? Also, you aren't required to use the Poetry build/publish/whatever steps to build wheels and distribute packages, the `python -m build --sdist` in the workflow file still works fine with these changes.

greglucas: Exactly, there is a build dependence on poetry. Granted there is a build dependence on setuptools in the current version (but I'd be fine moving to hatch, or any other build backend that more closely follows the PEPs with the `[project]` section instead of `[tool.poetry.project]`.

I guess you really shouldn't need to use poetry in the github actions at all because pip is already available as you described and will look at the `pyproject.toml`. I think you'd be able to change `pyproject.toml` over to using the poetry definitions of things, and then not touch `tests.yml` and it should work (avoid the poetry install step). (how I run everything locally already)

I will reiterate though that I am torn on this. I dislike poetry, but like consistency :) Hence why I think this should be a separate PR so others can approve it to get in!

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+                self.startdate,
+                self.enddate,
+                self.version,
+                self.extension,
+            ]
+        ):
+            error_message = (
+                f"Invalid filename, missing attribute. Filename "
+                f"convention is {FILENAME_CONVENTION} \n"
+            )
```
bourque, line 193: This is a neat way to do it!

tests/test_file_validation.py
bourque, line 1: Nice test coverage here!

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+}
+
+VALID_DATALEVELS = {"l0", "l1", "l1a", "l1b", "l1c", "l1d", "l2"}
+
+VALID_FILE_EXTENSION = {"pkts", "cdf"}
+
+FILENAME_CONVENTION = (
+    "<mission>_<instrument>_<datalevel>_<descriptor>_"
+    "<startdate>_<enddate>_<version>.<extension>"
+)
```
bourque, line 31: This is nitpicky and I don't have strong opinions on this, but I wonder if these should be defined in ``__init__.py`` so that other repos can import like this

```
import imap_data_access
imap_data_access.VALID_INSTRUMENTS
```

instead of 

```
from imap_data_access import file_validation
file_validation.VALID_INSTRUMENTS
```

I can see a case where a developer just needs a list of valid instruments without having to validate a filename.

bourque (approved):
Nice work, I like having a 'one-stop-shop for all of your file validation needs'

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,258 @@
+"""Methods for managing and validating filenames and filepaths"""
```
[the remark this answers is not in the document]
maxinelasp: Examples in the README is a good idea!

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+    "idex",
+    "lo",
+    "mag",
+    "swapi",
+    "swe",
+    "ultra-45",
+    "ultra-90",
+}
+
+VALID_DATALEVELS = {"l0", "l1", "l1a", "l1b", "l1c", "l1d", "l2"}
```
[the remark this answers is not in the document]
maxinelasp: Good catch, thanks!

imap_data_access/file_validation.py
maxinelasp: Yeah, I like that. I'll move them. The easier it is to access this stuff, the less likely it is that someone will re-define it, lol

imap_data_access/file_validation.py
maxinelasp: The main issue is that this is combining two ways of solving this from sds-data-manager. The functions are from a different file than the class. I was having a lot of trouble deciding which way would be better, if functions should be static inside the class or stand alone, etc. What are people's thoughts on a class vs standalone functions? This code would be used in different ways, so sometimes the class is better and sometimes the standalone methods are better.

imap_data_access/file_validation.py
maxinelasp: It was all Tenzin! I slightly modified the method but it was her idea.

imap_data_access/__init__.py
```diff
@@ -26,3 +27,28 @@
...
+    "idex",
+    "lo",
+    "mag",
+    "swapi",
+    "swe",
+    "ultra-45",
+    "ultra-90",
+}
+
+VALID_DATALEVELS = {"l0", "l1", "l1a", "l1b", "l1c", "l1d", "l2"}
```
laspsandoval, line 47: Sorry for the late comment. Ultra will also have: l1ca and l1cb.

laspsandoval (approved):
I had a single comment about Ultra data level. Thanks for the work!

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Returns
+    -------
+    Path
+        File path generated from the filename.
+
+    """
+
+    components = extract_filename_components(filename)
+
```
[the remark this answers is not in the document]
tech3371: Just checking if this is going to get combined or not in this PR?

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,264 @@
...
+VALID_FILE_EXTENSION = {"pkts", "cdf"}
+
+FILENAME_CONVENTION = (
+    "<mission>_<instrument>_<datalevel>_<descriptor>_"
+    "<startdate>_<enddate>_<version>.<extension>"
+)
+
+
+def extract_filename_components(filename: str | Path):
```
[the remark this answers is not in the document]
tech3371: is this going to be a staticmethod in the class now?

tech3371 (commented):
I had couple follow-up questions from today's discussion. Besides that, it looks good.

pyproject.toml
maxinelasp: I also dislike poetry, but I really hate setuptools 😆. 

Poetry is simple to use (IMO) but has some annoying opinions on virtual environments and also doesn't follow some recent PEPs. It's also kind of annoyingly opinionated. I find setuptools to be difficult to use and it doesn't have the concept of a lockfile, so version management is a lot more complex, and I find it's missing some key features. Also I hate the setuptools documentation. Hatch has only been around a few years - I'm worried it's not mature enough to commit to, although it is interesting. There's pipenv, but that also has some significant drawbacks and kind of seems generally bad. 

PDM seems like a good alternative, in that it aligns with that newish PEP about `[project]`, but you still need to install another dependency, and I'm not sure it's worth transitioning to something else when Poetry seems to be working pretty well for most people on the team. It's also vaguely newer, although it seems more commonly used in production systems than Hatch. If we were to switch, that's the one I would argue for, but I don't think it's so much better that we should switch. 

In conclusion, I don't know why this is still an apparently unsolved problem in Python but it is. Also, all these dependency systems, including setuptools involve a build dependency. I also wish that Pip and standard python just like... had these tools built in... but as you say, if you don't want to use poetry in the workflow file, you can just use Pip to install/build.

imap_data_access/file_validation.py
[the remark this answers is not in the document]
maxinelasp: Yes, this would be a static method in the class.

imap_data_access/file_validation.py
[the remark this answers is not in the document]
maxinelasp: It will be combined, I mostly just confused myself and it should be combined whether we go the class route or not.

imap_data_access/file_validation.py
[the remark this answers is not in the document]
maxinelasp: Oops forgot to do this in my last commit, will do now

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,226 @@
...
+
+import re
+from datetime import datetime
+from pathlib import Path
+
+import imap_data_access
+
+
+class ScienceFilepath:
+    class InvalidScienceFileError(Exception):
```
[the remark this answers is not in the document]
maxinelasp: Since this error is so tightly related to the class, I thought it might make sense to package them together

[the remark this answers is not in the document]
greglucas: I'm only 50-50 on this. I agree they are coupled, but now in other code I can't import the exception directly but instead have to access it from the `ScienceFilePath` class. I'm not sure the `raise self.InvalidScienceFileError()` gains anything here?

```python
from imap_data_access import file_validation

try:
    ...
except file_validation.ScienceFilepath.InvalidScienceFileError:

# versus
except file_validation.InvalidScienceFileError:
```

I'm not opposed to this way, so non-blocking comment.
Edit: I see that later on you directly import `ScienceFilepath` which reduces a bit of the verbosity from the example above.

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,226 @@
...
+            Format - YYYYMMDD.
+        <version>: This stores software version and data version. Version format is
+            vxx-xx.
+
+        Parameters
+        ----------
+        filename : str | Path
+            Science data filename or file path.
+        data_dir : Path, optional
+            Optional higher directory level for the data, by default None
```
greglucas, line 45: What do we mean by this variable? I think we only want to go through `imap_data_access.config["DATA_DIR"]` rather than have another option as well for multiple nestings.

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,226 @@
...
+        self.data_level = split_filename["datalevel"]
+        self.descriptor = split_filename["descriptor"]
+        self.startdate = split_filename["startdate"]
+        self.enddate = split_filename["enddate"]
+        self.version = split_filename["version"]
+        self.extension = split_filename["extension"]
+
+        self.error_message = self.validate_filename()
+        if self.error_message:
+            raise self.InvalidScienceFileError(f"{self.error_message}")
```
greglucas, line 69: Is there a reason to keep the error message on the class, or could we just raise from the end of the `validate_filename()` helper function directly instead?

```suggestion
        self.validate_filename()
```

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,226 @@
...
+            Path of dependency data.
+
+        Returns
+        -------
+        components : dict
+            Dictionary containing components.
+
+        """
+        pattern = (
+            r"^(?P<mission>imap)_"
```
greglucas, line 206: Do you want to take in whatever is passed and not restrict it to imap? Either way is fine I think, so feel free to ignore.
```suggestion
            r"^(?P<mission>[^_]+)_"
```

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,226 @@
...
+        """Class to store filepath and file management methods for science files.
+
+        If you have an instance of this class, you can be confident you have a valid
+        science file and generate paths in the correct format.
+
+        Current filename convention:
+        <mission>_<instrument>_<datalevel>_<descriptor>_<startdate>_<enddate>
+        _<version>.<extension>
+
+        NOTE: There are no optional parameters anymore. All parameters are required.
```
greglucas, line 27:
```suggestion
        NOTE: There are no optional parameters. All parameters are required.
```

imap_data_access/__init__.py
```diff
@@ -7,9 +7,10 @@
 import os
 from pathlib import Path
 
+from imap_data_access.file_validation import ScienceFilepath
 from imap_data_access.io import download, query, upload
 
-__all__ = ["query", "download", "upload"]
+__all__ = ["query", "download", "upload", "ScienceFilepath"]
```
greglucas, line 13: Is this something we want in the top-level namespace? Versus having this imported from the `file_validation` module? I don't have a strong preference either way, so just something to consider.

imap_data_access/file_validation.py
```diff
@@ -0,0 +1,226 @@
...
+from __future__ import annotations
+
+import re
+from datetime import datetime
+from pathlib import Path
+
+import imap_data_access
+
+
+class ScienceFilepath:
```
greglucas, line 11: I was expecting Path to be capitalized because Path is from pathlib... I'm not sure there is a standard for it other than I think in other areas of the codebase we are using `file_path` with a separator indicting two words. Non-blocking suggestion though.
```suggestion
class ScienceFilePath:
```

greglucas (approved):

imap_data_access/file_validation.py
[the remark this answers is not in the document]
maxinelasp: I'm not sure other files ever would raise that error, which is why I moved it into the class. It's not something majorly important to me either, mostly just that the two things are really closely coupled right now so it made sense to me to make that explicit. If this is an error that other code would raise then I agree it should be outside the class

imap_data_access/file_validation.py
maxinelasp: Mostly I wanted to move the work out of io.py into here, it made sense to add it as a keyword argument instead of getting it from the data directory just to have more flexibility for people importing the class and using it directly rather than using it through imap_data_access io. The imap_data_access methods pass in the data dir config to the class, but in imap_processing or something I think it makes sense to keep it only in the class rather than reading it from an environment variable.

maxinelasp: I can remove this if preferred, I thought it was a little cleaner than handling it in `io.py` but I can go back to handling it there if this is confusing

imap_data_access/file_validation.py
maxinelasp: I could see people using validate_filename without wanting to throw an error - when I see `validate_something` I wouldn't assume it throws an error if it's invalid.

imap_data_access/file_validation.py
maxinelasp: Yeah I wasn't sure which way was best, I just stuck with imap to not change the behavior

imap_data_access/file_validation.py
maxinelasp: Sure, technically "file path" is two words so this doesn't line up with naming conventions anyway, I will change

imap_data_access/__init__.py
maxinelasp: 🤷  I don't have a preference either, this is slightly cleaner since file_validation only has the `ScienceFilepath` class in it anyway

README.md
```diff
@@ -148,3 +148,26 @@ continue to encounter this, reach out to the IMAP SDC at
 This could mean that the local data directory is not set
 up with the same paths as the SDC. See the [data directory](#data-directory)
 section for an example of how to set this up.
+
+## File Validation
+
+This package validates filenames and paths to check they follow our standards, as defined by the filename conventions. There is also a class available for
+use by other packages to create filepaths and filenames that follow the IMAP SDC conventions.
+
+To use this class, use `imap_data_access.ScienceFilepath`.
```
greglucas, line 157:
```suggestion
To use this class, use `imap_data_access.ScienceFilePath`.
```

imap_data_access/file_validation.py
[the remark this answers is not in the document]
greglucas: :+1: Sounds good. It is fine to leave here then. I assume we'll have a `CalibrationFileError` on the calibration file path if we do add that, or can re-evaluate this then and pull it out later.

imap_data_access/file_validation.py
greglucas: I was torn as well allowing an extra argument in the `io.py` routines, so I think we're thinking the same thing here :) and there is not a right/wrong answer.

My reason for not adding it was to try and start locking down potential miscues and only make one place to change things, which would be the `config` dictionary and if you want to update that it could be through the command line, environment variables, or via attribute updating within the package.

From your example in the README:
```python
# Use a data directory
imap_data_access.config["DATA_DIR"] = Path("/data/personal")
science_file = imap_data_access.ScienceFilePath("imap_swe_l0_sci_20240105_20240105_v00-05.pkts")

# Filepath = /data/personal/imap/swe/l0/2024/01/imap_swe_l0_sci_20240105_20240105_v00-05.pkts
filepath = science_file.construct_file_path()

# or have the using routines construct it themselves?
filepath = imap_data_access.config["DATA_DIR"] / science_file.construct_file_path()
```

The two options above would return either the fully qualified path directly, or have the user construct it themselves if it kept returning the current relative path. My preference would probably be the second version, and removing the `data_dir` argument option and only allowing filename.

imap_data_access/file_validation.py
greglucas: I agree, but I also wouldn't expect an empty string or for an `error_message` attribute, so there are a few indirections here too.

It isn't a static method, so I think it'd only get used this one time, or are there places we'd want to use it after the fact too? Could make this private and call it `_raise_if_bad_filename()`.

PR #17, opened 2024-02-13
Minor bugfix to release workflow

maxinelasp:
# Change Summary
I noticed the release workflow would run both jobs at the same time, when the pypi publish step has a dependency on the release step. 

## Testing

Checked this on my fork: https://github.com/maxinelasp/imap-data-access/actions/runs/7892399492


tech3371 (approved):

bourque (approved):
Nice catch!

PR #20, opened 2024-02-19
Release version

maxinelasp:


imap_data_access/__init__.py
```diff
@@ -20,7 +20,7 @@
     "VALID_FILE_EXTENSION",
     "FILENAME_CONVENTION",
 ]
-__version__ = "0.3.1"
+__version__ = "0.4.0"
```
bourque, line 23: Doesn't really have anything to do with this PR, but we should probably decide what a `v1.0` release means and when we should try to do that.

bourque (approved):
LGTM

imap_data_access/__init__.py
maxinelasp: Agreed, I think once active development slows down on the file_validation stuff and we have definitely finalized the imap-data-access CLI we can do a 1.0 release. Main thing is that after that point we should really really avoid changing the CLI/access patterns without good reason, but having used it a fair amount for the imap_processing CLI updates, I think what we have now works well so we're close to that point, but definitely we should be 100% final with the CLI and the data API first.

PR #21, opened 2024-02-21
Consistent ruff checks with imap_processing

bourque:
# Change Summary

## Overview

Akin to https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/258, this PR adds the `D` and `S` `ruff` checks in order to be more consistent with `imap_processing` and `sds_data_manager`. I also made the corresponding fixes.

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -58,6 +58,7 @@ def download(file_path: Union[Path, str]) -> Path:
     -------
     pathlib.Path
         Path to the downloaded file
+
     """
     destination = imap_data_access.config["DATA_DIR"]
     if isinstance(file_path, str) and "/" not in file_path:
@@ -133,6 +134,7 @@ def query(
     -------
     list
         List of files matching the query
+
     """
     # locals() gives us the keyword arguments passed to the function
     # and allows us to filter out the None values
@@ -162,6 +164,7 @@ def upload(file_path: Union[Path, str]) -> None:
     file_path : pathlib.Path or str
         Path to the file to upload. It must be located within
         the ``imap_data_access.config["DATA_DIR"]`` directory.
+
     """
     file_path = Path(file_path).resolve()
     if not file_path.exists():
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -58,7 +58,6 @@ def download(file_path: Union[Path, str]) -> Path:
     -------
     pathlib.Path
         Path to the downloaded file
-
     """
     destination = imap_data_access.config["DATA_DIR"]
     if isinstance(file_path, str) and "/" not in file_path:
@@ -134,7 +133,6 @@ def query(
     -------
     list
         List of files matching the query
-
     """
     # locals() gives us the keyword arguments passed to the function
     # and allows us to filter out the None values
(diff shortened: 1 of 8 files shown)


imap_data_access/io.py
```diff
@@ -162,6 +164,7 @@ def upload(file_path: Union[Path, str]) -> None:
     file_path : pathlib.Path or str
         Path to the file to upload. It must be located within
         the ``imap_data_access.config["DATA_DIR"]`` directory.
+
```
greglucas, line 167: These newlines seem a little superfluous to me because they are still within the docstring 🤷 , I feel like I'd almost prefer the newline after this if anything.

pyproject.toml
```diff
@@ -48,5 +48,10 @@ addopts = "-ra"
 
 [tool.ruff]
 target-version = "py39"
-select = ["B", "E", "F", "I", "N", "W", "PL", "PT", "UP", "RUF"]
-ignore = ["PLR2004"]
+lint.select = ["B", "E", "D", "F", "I", "N", "S", "W", "PL", "PT", "UP", "RUF"]
+# D104: Missing docstring in public package
+# D203: 1 blank line required before class docstring
+# D213: Multi-line docstring summary should start at the second line
+# S101: Use of assert detected
```
greglucas, line 55: Should this only be in the tests directly under a `per-file-ignore` regex? I don't think we want assert, False, "message" in the main code and would prefer raising an exception instead.

tests/test_io.py
```diff
@@ -166,6 +256,16 @@ def test_upload_not_relative_to_base(monkeypatch, mock_urlopen):
     "upload_file_path", ["a/b/test-file.txt", Path("a/b/test-file.txt")]
 )
 def test_upload(mock_urlopen, upload_file_path):
+    """Test a basic call to the upload API.
```
greglucas, line 259: I am somewhat hesitant to add these docstring checks to the testing directories. I could see people writing fewer tests because they have to go through this with every new test they write and just don't want to deal with the hassle. I guess most tests usually don't take parameters in, so hopefully not too big of a deal and we can always remove the checks later if this does turn out to negatively impact test writing.

greglucas (approved):

pyproject.toml
bourque: Good point, I do think that is the better way to go on this one. I will update that.

tests/test_io.py
bourque: Yeah, I can see your point there. And usually test function names are written in a way that makes it obvious what is being tested.

Having _some_ docstring there does make the code reference docs a bit nicer looking, if we care about that.

imap_data_access/io.py
greglucas: :joy: I mean I'd personally prefer that, so if you do too I think we should. We don't have to take every check, I think it is just getting it to something you like. I'm also fine leaving it in, it just stuck out to me here.

tests/test_io.py
greglucas: I think we leave it in as you have it since you've done the work :) Then we can adjust the lint codes later if it is a problem.

PR #22, opened 2024-02-23
FIX: Update the filename conventions and parsing

greglucas:
# Change Summary

## Overview
This follows the new file naming convention that the SPDF would like to follow. Changing some underscores to hyphens and adding an optional repointing field.

There is now a 4-digit repointing field that is `None` if it isn't present. Do we want to add `rp` or some other clarifier before that field to indicate what it is?

### DIFF
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -42,15 +42,13 @@ VALID_INSTRUMENTS = {
     "codice",
     "glows",
     "hit",
-    "hi-45",
-    "hi-90",
+    "hi",
     "idex",
     "lo",
     "mag",
     "swapi",
     "swe",
-    "ultra-45",
-    "ultra-90",
+    "ultra",
 }
 
 VALID_DATALEVELS = {
@@ -60,7 +58,8 @@ VALID_DATALEVELS = {
     "l1b",
     "l1c",
     "l1ca",
-    "l1cb" "l1d",
+    "l1cb",
+    "l1d",
     "l2",
     "l2pre",
     "l3",
@@ -74,5 +73,5 @@ VALID_FILE_EXTENSION = {"pkts", "cdf"}
 
 FILENAME_CONVENTION = (
     "<mission>_<instrument>_<datalevel>_<descriptor>_"
-    "<startdate>_<enddate>_<version>.<extension>"
+    "<startdate>(-<repointing>)_<version>.<extension>"
 )
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -49,7 +49,7 @@ def _print_query_results_table(query_results: list[dict]):
     print(f"Found [{num_files}] matching files")
     if num_files == 0:
         return
-    format_string = "{:<10}|{:<10}|{:<10}|{:<10}|{:<8}|{:<7}|{:<50}|"
+    format_string = "{:<10}|{:<10}|{:<10}|{:<10}|{:<10}|{:<7}|{:<50}|"
     # Add hyphens for a separator between header and data
     hyphens = "-" * 111 + "|"
     print(hyphens)
@@ -58,7 +58,7 @@ def _print_query_results_table(query_results: list[dict]):
         "Data Level",
         "Descriptor",
         "Start Date",
-        "End Date",
+        "Repointing",
         "Version",
         "Filename",
     ]
@@ -72,7 +72,7 @@ def _print_query_results_table(query_results: list[dict]):
             item["data_level"],
             item["descriptor"],
             item["start_date"],
-            item["end_date"],
+            item["repointing"],
             item["version"],
             os.path.basename(item["file_path"]),
         ]
@@ -95,7 +95,7 @@ def _query_parser(args: argparse.Namespace):
         "data_level",
         "descriptor",
         "start_date",
-        "end_date",
+        "repointing",
         "version",
         "extension",
     ]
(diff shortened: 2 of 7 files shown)


bourque (approved):
This looks good to me (if this is indeed the filename convention we can/should use ... it seems like there still might be some changes afoot from recent emails from Robert)

imap_data_access/__init__.py
```diff
@@ -73,6 +73,6 @@
 VALID_FILE_EXTENSION = {"pkts", "cdf"}
```
tech3371, line 73: can we update instruments name too?
```
VALID_INSTRUMENTS = {
    "codice",
    "glows",
    "hit",
    "hi",
    "idex",
    "lo",
    "mag",
    "swapi",
    "swe",
    "ultra",
}
```

tech3371: Also, is there a command missing between "l1cb" and "l1d"?

imap_data_access/__init__.py
```diff
@@ -73,6 +73,6 @@
...
 
 FILENAME_CONVENTION = (
-    "<mission>_<instrument>_<datalevel>_<descriptor>_"
-    "<startdate>_<enddate>_<version>.<extension>"
+    "<mission>_<instrument>_<datalevel>-<descriptor>_"
+    "<startdate>-<enddate>(-<repointing>?)_<version>.<extension>"
```
tech3371, line 77:
```suggestion
    "<mission>_<instrument>_<datalevel>_<descriptor>_"
    "<startdate>(-<repointing>)_<version>.<extension>"
```

imap_data_access/file_validation.py
```diff
@@ -26,21 +26,19 @@ def __init__(self, filename: str | Path):
...
             decided by each instrument. For L0, "raw" is used.
-        <startdate>: startdate is the earliest date in the data. Format - YYYYMMDD
-        <enddate>: Some instrument and some data level requires to store date range.
-            If there is no end date, then startdate will be used as enddate as well.
-            Format - YYYYMMDD.
-        <version>: This stores software version and data version. Version format is
-            vxx-xx.
+        <start_date>: startdate is the earliest date in the data, format: YYYYMMDD
+        <repointing?>: This is an optional field. It is used to indicate which
+            repointing the data is from, format: XXXXX
```
tech3371, line 40:
```suggestion
        <repointing>: This is an optional field. It is used to indicate which
            repointing the data is from. Repointing format is repoint followed by 5-digits: repointXXXXX
```

tech3371: We had a discussion on Slack about prefixing repoint to give it some meaning like how version number starts with `vxxx`. You mentioned about doing something like this for repointing, `repointXXXXX`. Do we still want to do this? That's why I suggested above change.

imap_data_access/file_validation.py
```diff
@@ -254,11 +252,11 @@ def extract_filename_components(filename: str | Path) -> dict:
         pattern = (
             r"^(?P<mission>imap)_"
             r"(?P<instrument>[^_]+)_"
-            r"(?P<data_level>[^_]+)_"
+            r"(?P<data_level>[^-]+)_"
```
tech3371, line 255: should we undo this change based on recent discussion?

imap_data_access/file_validation.py
```diff
@@ -254,11 +252,11 @@ def extract_filename_components(filename: str | Path) -> dict:
...
             r"(?P<descriptor>[^_]+)_"
-            r"(?P<start_date>\d{8})_"
-            r"(?P<end_date>\d{8})_"
-            r"(?P<version>v\d{2}-\d{2})"
+            r"(?P<start_date>\d{8})"
+            r"(-repoint(?P<repointing>\d{5}))?"  # Optional repointing field
```
tech3371, line 258: Oh cool. You are using prefix `repointXXXXX`

imap_data_access/__init__.py
greglucas: Good catch on the instruments, updated! I also updated the l1cb / l1d, which I think was a typo from before.

imap_data_access/file_validation.py
greglucas: I added the `repoint` qualifier, but left off the extra text in the format section.

imap_data_access/file_validation.py
greglucas: Yes, I think you're right, this should be underscore. I've updated it now!

imap_data_access/__init__.py
greglucas: Thanks for catching this, updated!

imap_data_access/cli.py
```diff
@@ -242,14 +242,17 @@ def main():
     query_parser.add_argument(
```
tech3371, line 242: should we remove this now?

imap_data_access/cli.py
```diff
@@ -242,14 +242,17 @@ def main():
...
         "--end-date", type=str, required=False, help="End date in YYYYMMDD format"
     )
+    query_parser.add_argument(
+        "--repointing", type=str, required=False, help="Repointing number"
```
tech3371, line 246:
```suggestion
        "--repointing", type=str, required=False, help="Repointing number. Eg. 'repointXXXXX'"
```

tech3371 (approved):
So far, it looks great to me. I will test it with infrastructure after you release it and open issue if I find any bugs. Thank you for doing this!

PR #32, opened 2024-04-18
REL: v0.5.1

greglucas:
Bump version to v0.5.1

Will release to GitHub/PyPI after this goes in. Let me know if you think I missed anything else we need to update, or if you think this should be a minor instead of a patch (I think it is only bugfix here).


maxinelasp (approved):
LGTM!

PR #33, opened 2024-05-03
ENH: Allow SPICE file downloads/uploads

### LINKED ISSUE #23: BUG - Download not working without full file path
### ## Description of the issue

The download method in the CLI doesn't work properly if a file is requested via filename (as opposed to full file path.) The CLI attempts to add IMAP_DATA_DIR to the front of the filename. 

1. 2. 3.

1. Run `imap-data-access download imap_mag_l1a_raw-norm_20240221_20240221_v01-00.cdf`

### Code Snippet:

```python
# in io.py
destination = imap_data_access.config["DATA_DIR"]
if isinstance(file_path, str) and "/" not in file_path:
    # Construct the directory structure from the filename
    # This is for science files that contain the directory structure in the filename
    # Otherwise, we assume the full path to the file was given

    # ERROR IS HERE: file_path is never updated with full expected path
    science_file_path = imap_data_access.ScienceFilePath(file_path)

    destination = science_file_path.construct_path()
else:
    destination /= file_path
```

File should be able to be downloaded from just the filename

404 Error

### Additional notes

_No response_

### Affected areas (code, data, or process)

io.py

### Suggested fix?

set `file_path = science_file_path.construct_filepath()`

greglucas:
# Change Summary

## Overview

This adds the ability to download and upload SPICE files to the SDC. There is a mapping based on the SPICE file extension to which subdirectory it will go to.

Also closes #23 because we are now always inferring the directory structure from the filename, even if a full path was given to the file.

```bash
# This for the new SPICE file
imap-data-access -v download imap_001.tf
# and this for issue 23
imap-data-access -v download imap_mag_l1a_burst-magi_20240314_v001.cdf
```

## Testing

I added a test_science_file and test_science_path now because we require explicit science file paths or SPICE paths for download/upload now, so this makes it easier to re-use during tests.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -276,3 +276,61 @@ class ScienceFilePath:
             # We want the repointing number as an integer
             components["repointing"] = int(components["repointing"])
         return components
+
+
+class SPICEFilePath:
+    """Class for building and validating filepaths for SPICE files."""
+
+    class InvalidSPICEFileError(Exception):
+        """Indicates a bad file type."""
+
+        pass
+
+    def __init__(self, filename: str | Path):
+        """Class to store filepath and file management methods for SPICE files.
+
+        If you have an instance of this class, you can be confident you have a valid
+        SPICE file and generate paths in the correct format. The parent of the file
+        path is set by the "IMAP_DATA_DIR" environment variable, or defaults to "data/"
+
+        IMAP_DATA_DIR/spice/<subdir>/filename"
+
+        Parameters
+        ----------
+        filename : str | Path
+            SPICE data filename or file path.
+        """
+        self.filename = Path(filename)
+        self.spice_dir = imap_data_access.config["DATA_DIR"] / "imap/spice"
+
+        if self.filename.suffix[1:] not in imap_data_access.VALID_SPICE_EXTENSIONS:
+            raise self.InvalidSPICEFileError(
+                f"Invalid SPICE file. Expected file to have one of the following "
+                f"extensions {imap_data_access.VALID_SPICE_EXTENSIONS}"
+            )
+
+    def construct_path(self) -> Path:
+        """Construct valid path from the class variables and data_dir.
+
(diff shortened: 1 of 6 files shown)


imap_data_access/file_validation.py
```diff
@@ -276,3 +276,61 @@ def extract_filename_components(filename: str | Path) -> dict:
...
+        suffix_mapping = {
+            ".bc": "ck",
+            ".tf": "fk",
+            ".tls": "lsk",
+            ".tm": "mk",
+            ".tpc": "pck",
+            ".tsc": "sclk",
+            ".bsp": "spk",
+        }
+        subdir = suffix_mapping[self.filename.suffix]
```
greglucas, line 333: What do we want to do for the suffixes that are in `VALID_SPICE_EXTENSIONS`, but not in the `suffix_mapping`? Right now, this will fail here because it isn't in the mapping. Do we want to put them at a default level, or only consider these the valid suffixes?

imap_data_access/io.py
```diff
@@ -61,16 +61,19 @@ def download(file_path: Union[Path, str]) -> Path:
...
+        path_obj = imap_data_access.ScienceFilePath(file_path.name)
 
-        science_file_path = imap_data_access.ScienceFilePath(file_path)
+    destination = path_obj.construct_path()
 
-        destination = science_file_path.construct_path()
-    else:
-        destination /= file_path
+    # Update the file_path with the full path for the download below
+    file_path = str(destination.relative_to(imap_data_access.config["DATA_DIR"]))
```
greglucas, line 76: This is the key to be able to download a file with either the filename, or the full path. Make sure we make it a full relative path before sending the request.

subagonsouth (commented):
Other than the mismatch that you pointed out between the `VALID_SPICE_EXTENSIONS` and the mapping to spice directories, this looks good to me.

subagonsouth (approved):
This looks great.

PR #34, opened 2024-05-03
Add a way to add an api-key authentication to the upload request

greglucas:
This adds an X-api-key header to the upload request. The api key can be set as an environment variable, through the command line, or through the config dictionary.


maxinelasp (approved):
I like it, is the general idea to get this API key from cognito? Good to see some progress on security stuff though

greglucas:
> I like it, is the general idea to get this API key from cognito? Good to see some progress on security stuff though

Thats a good question. My initial thought was no and we would generate these for people and send them their API token (I think this means only the POC and us as developers [we can make our own, so really one external entity is all]). We would generate these through the CLI, which I'll add documentation to in the other repository. We could add Cognito here instead and use that as the auth layer, but we'd still want to control who has access to create a Cognito identity.

Thoughts on whether generating tokens through the CLI is good enough or if we want to add in some form of Cognito auth?

maxinelasp:
Tokens through the CLI seems fine to me!

PR #37, opened 2024-05-07
Fix api key used for uploads

greglucas:
# Change Summary

## Overview

I messed up, and didn't fully test the previous version I guess. There was an issue when NOT using an API-key, and we want users to be able to go through either route and let the server complain not our application.

There was an error converting None-type, so we need to explicitly make it an empty dictionary rather than adding {"Header-value": None} to the request.

I have added a more explicit test for the None/expected-string case now.

I have tested this on the dev server and all uploads/downloads worked for both science and spice files.


bourque (approved):
LGTM

PR #43, opened 2024-06-10
FIX: Remove restriction on upload data directory structure

### LINKED ISSUE #41: Remove the requirement to upload from a specific data folder structure
### Summary

It would be nice to have the ability to upload data files from any directory structure and not require the command to be run from a specific current working directory + data/ folder combination. It is currently confusing where you need to run the cli from, so removing all folder path requirements on the upload would be helpful.

cc @vmartinez-cu @tech3371, feel free to add clarifications to this that would help

### Requirements

The more specific the better.
  - [ ] Remove upload folder path requirement
  - [ ] Should be able to take in full path to folder `/root/path/to/file.txt` and `path/to/file.txt` (if you're currently in `/root`)

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview

Our APIs don't require the full path anymore, they can use just the filename and infer the directory structure. This is helpful for the upload to be able to upload a file from the current working directory.

```python
imap-data-access -v upload imap_mag_l1a_burst-magi_20240314_v007.cdf 
INFO:imap_data_access.io:Uploading file /Users/grlu5547/code/imap/imap-data-access/data/imap/mag/l1a/2024/03/imap_mag_l1a_burst-magi_20240314_v007.cdf to https://api.dev.imap-mission.com/upload/imap_mag_l1a_burst-magi_20240314_v007.cdf
HTTP Error: 409 - Conflict
Server Message: "imap/mag/l1a/2024/03/imap_mag_l1a_burst-magi_20240314_v007.cdf already exists."
```

closes #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ The folder structure for data files within the IMAP SDC is rigidly
 defined, so the data access will mimic that structure to make sure
 all data is stored in the same heirarchical structure as the SDC.
 This will enable seamless transition between a user's local system
-and the SDC.
+and the SDC. This is only used for downloads.
 
 A user's root data location can be specified as an environment
 variable ``IMAP_DATA_DIR`` or through a configuration dictionary
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -164,8 +164,7 @@ def upload(file_path: Union[Path, str], *, api_key: Optional[str] = None) -> Non
     Parameters
     ----------
     file_path : pathlib.Path or str
-        Path to the file to upload. It must be located within
-        the ``imap_data_access.config["DATA_DIR"]`` directory.
+        Path to the file to upload.
     api_key : str, optional
         API key to authenticate with the data access API. If not provided,
         the value from the IMAP_API_KEY environment variable will be used.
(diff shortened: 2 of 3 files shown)


tech3371:
I was trying to find doc strings to update which reflects this change. Looks like there are few places to update. I can help pin down which ones needs update.

greglucas:
> I was trying to find doc strings to update which reflects this change. Looks like there are few places to update. I can help pin down which ones needs update.

I didn't find any that needed to be updated, so if you have ones you think should be changed then yes please let me know. I thought all of the other locations were valid still. For example: https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/82e9bbb1b51c30a41c10af6d6892747ded89e0de/imap_data_access/cli.py#L169-L170
is still correct because you must give the path to the file, we don't infer it anywhere. Maybe changing it to "This can be a relative or absolute path." would help though?

greglucas:
@tech3371 I didn't actually remove the default directory `data/` requirement on anything other than uploading. So, all path validation and setting of the `IMAP_DATA_ACCESS` global variables are still valid I think. The only thing this PR does is remove the check that we are within that structure. It lets you upload relative paths. `imap-data-access path/to/file.txt` is equivalent to `imap-data-access $PWD/path/to/file.txt` now and doesn't care that IMAP_DATA_ACCESS directory is pointing to `/my/other/location`.

vmartinez-cu:
> > I was trying to find doc strings to update which reflects this change. Looks like there are few places to update. I can help pin down which ones needs update.
> 
> I didn't find any that needed to be updated, so if you have ones you think should be changed then yes please let me know. I thought all of the other locations were valid still. For example:
> 
> https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/82e9bbb1b51c30a41c10af6d6892747ded89e0de/imap_data_access/cli.py#L169-L170
> 
> 
> is still correct because you must give the path to the file, we don't infer it anywhere. Maybe changing it to "This can be a relative or absolute path." would help though?

The README has the following statement regarding the data directory. Does this only apply to downloading files and not to uploading files?

```If the ``IMAP_DATA_DIR`` variable is not set, the program defaults
to the user's current working directory + ``data/``.```

vmartinez-cu (approved):

greglucas:
> The README has the following statement regarding the data directory. Does this only apply to downloading files and not to uploading files?

> If the ``IMAP_DATA_DIR`` variable is not set, the program defaults to the user's current working directory + ``data/``.

Yes, that is correct with this PR. Would it help to add something like this sentence?
> The ``IMAP_DATA_DIR`` is only used for downloads.

vmartinez-cu:
> Yes, that is correct with this PR. Would it help to add something like this sentence?
> 
> > The `IMAP_DATA_DIR` is only used for downloads.

Okay thanks for the clarification. Yeah, I think a short note about that would be helpful

greglucas:
Merging now, feel free to open another PR or leave suggestions here and I'm happy to follow-up if there are any other docs suggestions.

PR #44, opened 2024-06-11
FIX: Provide non-zero exit code when an action fails

greglucas:
# Change Summary

## Overview

We were catching the exception and printing the message to format the string nicer and not raise errors from the cli. But, when testing with another command line program we weren't getting a known error due to the zero exit code. This moves the handling of the try/except up into the parser and just directly exits with a non-zero exit code.

## Testing

```bash
# Now no longer a Python traceback with the ValueError
imap-data-access --data-dir /a/b/c upload a/b/c

# Now a 1 exit code (file doesn't exist)
imap-data-access download imap_codice_l1a_hskp_20240515_v001.cdf
```

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -31,11 +31,8 @@ def _download_parser(args: argparse.Namespace):
     args : argparse.Namespace
         An object containing the parsed arguments and their values
     """
-    try:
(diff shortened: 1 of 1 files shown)


tech3371 (approved):
same. LGTM

PR #63, opened 2024-07-10
Changes to API Cli -h  Outputs To Make Usage Easier

daralynnrhode:
# Mainly additions added to the argsparse outputs to make navigation from the command line easier.

## Overview

Added directions in the main `-h` menu to see `download/upload/query -h` menu for more information. These outputs will definitely be changed again once more utilities are added to the query, but for now if was helpful for me to see how the cli was operating. 

I anticipate many more changes to the api wrapper code later on.
## Updated Files

- cli.py
   - Changes to output statements and some additions 
   - added a try/except catch for when no params are given to cli.
- io.py
   - Changes to output statements when errors occur.


imap_data_access/cli.py
```diff
@@ -125,7 +126,7 @@ def _upload_parser(args: argparse.Namespace):
     print("Successfully uploaded the file to the IMAP SDC")
 
 
-def main():
+def main():  # noqa: PLR0915
```
bourque, line 129:
```suggestion
# PLR0915: too many statements
def main():  # noqa: PLR0915
```
Just to provide the reader with some context

imap_data_access/cli.py
```diff
@@ -147,18 +148,32 @@ def main():
     )
     download_help = (
         "Download a file from the IMAP SDC to the locally configured data directory. "
+        "Run 'download -h' for more information. "
+    )
+    help_menu_download = (
```
bourque, line 153:
```suggestion
    download_description = (
```

Since these strings (this one, `help_menu_query`, and `help_menu_upload`) are being used as descriptions in the `descriptions` parameter later on in the code, perhaps this is a better variable name? No strong opinions on this one.

bourque (changes requested):
I think we will want to remove the poetry lock file (assuming @greglucas agrees with me 😅 ). Otherwise, this looks good to me! These are very helpful changes.

poetry.lock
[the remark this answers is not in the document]
daralynnrhode: whoops. I think it accidentally got added because I am prone to use `git add --all` whenever I forget which file I am workin on.

imap_data_access/cli.py
daralynnrhode: I changed the name to be more specific as to where the statement occurs in the cli.

poetry.lock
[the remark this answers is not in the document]
maxinelasp: Agreed with Matthew, and sorry again for misleading you on the use of poetry in this project because I didn't check 😅  to avoid in the future, I like to use `git add -u` which only adds tracked updated files, or `git add -i` which allows for interactive adds (including partial adds if you have some code in the same file you don't want to commit!)

imap_data_access/cli.py
```diff
@@ -265,13 +282,18 @@ def main():
...
     parser_upload.add_argument("file_path", type=Path, help=file_path_help)
     parser_upload.set_defaults(func=_upload_parser)
 
     # Parse the arguments and set the values
-    args = parser.parse_args()
+    try:
+        args = parser.parse_args()
+    except Exception:
+        print("Please provide input parameters.")
+        sys.exit(0)
```
maxinelasp, line 295: Thanks for adding this! Two nitpicks - you should catch very specific exceptions (in this case, a typeerror is the one we're concerned about) and rather than a print statement and then an exit, you can just raise a custom exception. (you can see a list of exception types [here](https://docs.python.org/3/library/exceptions.html), a "valueerror" is a nice generic one for custom exceptions in many cases. It's best to pick an exception that makes sense if possible, or even write a custom exception.)

```suggestion
    except TypeError:
        raise ValueError("Please provide input parameters, or use `-h` for more information.")
```

So, this will still raise an error, which we want because the code is in a bad state and can't continue. But, it will be a super clear and simple error, with your custom text.  

More info on exceptions: https://realpython.com/python-raise-exception/

maxinelasp (commented):
Looks good to me once you adjust the try/except!

imap_data_access/cli.py
greglucas: For clis it is nicer to do parser.exit() so you don't get the full traceback and get a failing error code with the cli name etc.

See below for an example of that.

daralynnrhode: I see, I see. Let me know if what I changed it to is what is wanted and if it adequately conveys the problem raised.

greglucas (approved):
Looks good. Thanks for making that update. One further thing is whether the str(e) provides useful information or if your help string is good enough. I.e. can you remove the str(e) and just use your string help.

maxinelasp (approved):
Nice, looks good. I'd agree with Greg that you could probably remove the str(e) from ln 295, but it's up to you if you think that provides useful info to the end user.

bourque (approved):
Thanks for making those changes. Looks great!

PR #65, opened 2024-07-15
API Version Latest Query Return

### LINKED ISSUE #38: Add an option to the API to retrieve the latest version of a given file
### Summary

It would be nice to be able to pass in a "latest" flag or "latest" to the version, and have the query part of the API retrieve only the highest version number that matches the other requirements. This would be useful for ancillary files as well.

Note: This could be a good student ticket for exposure to the API and argparse.

### Requirements

The more specific the better.
  - [ ] Add a "latest" option to the "--version" input
  - [ ] Have the query with that flag only return the highest version numbers that match the rest of the requirements.

### Related tickets

_No response_

daralynnrhode:
# Adding a catch to the API wrapper to return latest version of a given file

## Overview

Made changes in the cli to catch if `--version latest` is passed in. Version would then need to be removed in order for a correct query to be returned. That query is then filtered based on what the greatest version number is. 

Closes #38 
## Updated Files

- cli.py

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -47,9 +47,9 @@ def _print_query_results_table(query_results: list[dict]):
     print(f"Found [{num_files}] matching files")
     if num_files == 0:
         return
-    format_string = "{:<10}|{:<10}|{:<10}|{:<10}|{:<10}|{:<7}|{:<50}|"
+    format_string = "| {:<10} | {:<10} | {:<15} | {:<10} | {:<10} | {:<7} | {:<50}|"
     # Add hyphens for a separator between header and data
-    hyphens = "-" * 113 + "|"
+    hyphens = "|" + "-" * 131 + "|"
     print(hyphens)
     header = [
         "Instrument",
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -125,7 +125,8 @@ def _upload_parser(args: argparse.Namespace):
     print("Successfully uploaded the file to the IMAP SDC")
 
 
-def main():
+# PLR0915: too many statements
+def main():  # noqa: PLR0915
     """Parse the command line arguments.
 
     Run the command line interface to the IMAP Data Access API.
@@ -147,6 +148,10 @@ def main():
     )
     download_help = (
         "Download a file from the IMAP SDC to the locally configured data directory. "
+        "Run 'download -h' for more information. "
+    )
+    help_menu_for_download = (
+        "Download a file from the IMAP SDC to the locally configured data directory. "
     )
     file_path_help = (
         "This must be the full path to the file."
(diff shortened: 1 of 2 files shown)


daralynnrhode:
I plan on doing a General Documentation update soon for the API wrapper. My hope is for it to be a clear outline of how to utilize the wrapper and what features it supports. That way certain commands like `--version latest` don't get coded and then never mentioned. My plan is to include everything that the api wrapper should be able to do and then label certain commands as WIP .

daralynnrhode:
https://github.com/all-contributors please add @daralynnrhode for code

daralynnrhode:
@all-contributors please add @daralynnrhode for doc

imap_data_access/cli.py
```diff
@@ -80,6 +80,11 @@ def _print_query_results_table(query_results: list[dict]):
     print(hyphens)
 
 
+def extract_version_number(str_version: str) -> int:
+    """Take a version string 'vXXX' and return and int 'XXX'."""
```
bourque, line 84: Could you please add a `Parameters` and `Returns` section here?

imap_data_access/cli.py
```diff
@@ -80,6 +80,11 @@ def _print_query_results_table(query_results: list[dict]):
     print(hyphens)
 
 
+def extract_version_number(str_version: str) -> int:
```
bourque, line 83:
```suggestion
def _extract_version_number(str_version: str) -> int:
```
Pretty nitpicky here but I think this function could be considered 'internal only' to this module, so by [PEP8 conventions](https://peps.python.org/pep-0008/#descriptive-naming-styles) should be named with a leading underscore

imap_data_access/cli.py
```diff
@@ -99,13 +104,32 @@ def _query_parser(args: argparse.Namespace):
...
+    # if latest version was included in search then filter returned query for largest.
+    if latest_version == "latest":
+        max_version = max(
+            extract_version_number(each_dict.get("version"))
+            for each_dict in query_results
+        )
+        query_results = [
+            each_dict
+            for each_dict in query_results
+            if extract_version_number(each_dict["version"]) >= max_version
```
bourque, line 130:
```suggestion
            if extract_version_number(each_dict["version"]) == max_version
```

I think logically this should be an exact match?

imap_data_access/cli.py
```diff
@@ -99,13 +104,32 @@ def _query_parser(args: argparse.Namespace):
...
+
+    # removing version from query if it is 'latest', updating flag
+    if args.version == "latest":
+        del query_params["version"]
+        latest_version = "latest"
+
     query_results = imap_data_access.query(**query_params)
 
+    # if latest version was included in search then filter returned query for largest.
+    if latest_version == "latest":
```
bourque, line 122:
```suggestion
    query_results = imap_data_access.query(**query_params)

    # if latest version was included in search then filter returned query for largest.
    if args.version == "latest":
```

I _think_ you could avoid having to define another variable here by just using `args.version` instead.

bourque (changes requested):
Nice work, I've played around with this locally and it is working really well!  I have some suggestions to simplify and rename a few things, otherwise this is looking good.

imap_data_access/cli.py
daralynnrhode: I agree with @bourque because the `max_version` is first assigned based on the max version found in the retrieved query. There is no way for any versions in the retrieved query to be greater than the `max_value`.

imap_data_access/cli.py
```diff
@@ -99,13 +104,32 @@ def _query_parser(args: argparse.Namespace):
...
+    latest_version = ""
     query_params = {
         key: value
         for key, value in vars(args).items()
         if key in valid_args and value is not None
     }
+
+    # removing version from query if it is 'latest', updating flag
+    if args.version == "latest":
+        del query_params["version"]
```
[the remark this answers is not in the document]
daralynnrhode: @maxine How would I do that? from what I can tell, `io.query()` does not validate if what is passed in is correct. I would assume checking the version would be similar to checking the instrument names passed in.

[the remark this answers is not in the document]
maxinelasp: Yeah, you know, you are right about that. However, this isn't really a validation step so much as an additional piece of functionality. I believe query doesn't validate because we assume that any code that is calling it is pre-validating everything already, whereas from the command line human error is more prone. 

I do kind of think all validation should be moved into the query() call, since anyone using the library should get that validation too. But, that should be something we discuss with the team/with Greg since he wrote a lot of this code. 

Since this is a feature add that code might want to use, I do feel it should be in `query` regardless of the other validation. What do you think?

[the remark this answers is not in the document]
daralynnrhode: I think I get what you are saying

maxinelasp (approved):
Thanks! The docstring for query() needs an update but after that it looks good!

bourque (approved):
Looks good, thanks for making those changes.

PR #69, opened 2024-07-16
 Query By Filename

### LINKED ISSUE #59: Data Access API Add Query Option for Filename
### Summary

Within the Query code, add an option to search for a file by the filename. See User Testing #40 for more information.

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

daralynnrhode:
# Adding a Query by filename argument

## Overview

Giving the option to query to see if a certain file based off of a given file path is present. Users would pass in `query --filename [File]` and the file will either be returned or no matching files will be returned. 
To avoid overwriting other passed in parameters, the filename must be the only argument passed in to the query.

Closes #59 
## Updated Files

- cli.py
   - added `--filename` to acceptable arguments
   - Code to pass the given file information into the query

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -47,9 +47,9 @@ def _print_query_results_table(query_results: list[dict]):
     print(f"Found [{num_files}] matching files")
     if num_files == 0:
         return
-    format_string = "{:<10}|{:<10}|{:<10}|{:<10}|{:<10}|{:<7}|{:<50}|"
+    format_string = "| {:<10} | {:<10} | {:<15} | {:<10} | {:<10} | {:<7} | {:<50}|"
     # Add hyphens for a separator between header and data
-    hyphens = "-" * 113 + "|"
+    hyphens = "|" + "-" * 131 + "|"
     print(hyphens)
     header = [
         "Instrument",
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -18,7 +18,10 @@ Use
 import argparse
 import logging
 import os
+import sys
 from pathlib import Path
+
+from file_validation import ScienceFilePath
 
 import imap_data_access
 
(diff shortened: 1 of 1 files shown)


imap_data_access/cli.py
```diff
@@ -98,14 +106,50 @@ def _query_parser(args: argparse.Namespace):
...
     query_params = {
         key: value
         for key, value in vars(args).items()
         if key in valid_args and value is not None
     }
+
+    # Checking to see if a filename was passed.
+    if args.filename is not None:
+        if len(sys.argv) > 4:  # I have doubts about this line
+            raise TypeError("Too many arguments, '--filename' should be ran by itself")
```
maxinelasp, line 121: To address your doubts - this seems ok to me. You could also check to see if "instrument", "data_level", etc is in sys.argv, which would be more reliable in case we add some additional options which _can_ be passed in alongside filename.

maxinelasp: Actually, you could also validate that `query_params` is empty, since we would expect it to be empty if `filename` was passed in, and we're rewriting it anyway so it's a good thing to check

maxinelasp (commented):
Looks like this PR does include commits from the other change to add "latest" - which is fine, but don't merge those changes until the other PR is approved (or remove them from here)

maxinelasp (approved):
Looks good!

imap_data_access/cli.py
```diff
@@ -277,6 +305,9 @@ def main():  # noqa: PLR0915
         choices=["table", "json"],
         default="table",
     )
+    query_parser.add_argument(
+        "--filename", type=str, required=False, help="Name of a file to be searched for"
```
bourque, line 309: Not absolutely necessary, but might be nice to add a link to the naming convention documentation in the help string here, so users can find what the file naming convention is in case they don't know it:

https://imap-processing.readthedocs.io/en/latest/development-guide/style-guide/naming-conventions.html#data-product-file-naming-conventions

bourque (approved):
LGTM! Nice work.

PR #76, opened 2024-08-07
Error when querying "latest" version when no data exists in database

mfacchinelli:
# Change Summary

## Overview
When querying the database with `version` == `latest`, but no matching data exists, Python errors. This change fixes the issue.

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
N/A

## Updated Files
- `io.py`
   - add check to make sure at least one data point exists before looking for latest version

## Testing
Tested locally with and without data

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -173,7 +173,7 @@ def query(
         logger.debug("Decoded JSON: %s", items)
 
     # if latest version was included in search then filter returned query for largest.
-    if (version == "latest") and (len(items) > 0):
+    if (version == "latest") and items:
         max_version = max(int(each_dict.get("version")[1:4]) for each_dict in items)
         items = [
             each_dict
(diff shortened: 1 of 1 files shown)


imap_data_access/io.py
```diff
@@ -173,7 +173,7 @@ def query(
         logger.debug("Decoded JSON: %s", items)
 
     # if latest version was included in search then filter returned query for largest.
-    if version == "latest":
+    if (version == "latest") and (len(items) > 0):
```
greglucas, line 176: As an FYI, you can also check against an empty list
```suggestion
    if (version == "latest") and items:
```

greglucas (approved):
It would be nice to get a test for the "latest" case here, but it looks like we don't currently have any tests for it, so we shouldn't necessarily put that on this simple update here since it would be a little involved I think to mock up the return.

One thing that would be nice is to codify what we mean by "latest", I would have actually assumed it was "latest version for each combination of other parameters", but it looks like it is actually doing "latest version for any combination of parameters"

i.e. `[(mag, l2, sci, v001), (swe, l2, sci, v002)]` would only return the swe file. It would be good to make sure that is the choice we made with tests and possibly even add some more documentation around which choice we selected there.

PR #79, opened 2024-08-15
Change For Dynamic Table Output

### LINKED ISSUE #53: Fix output table formatting to be Dynamic instead of Static
When using the `imap-data-access query <parameters>` option, the table will look off-kilter of the descriptors of the file are quite lengthy. Fix this format so the table looks tidy. 

To Note: Table has been updated, but could be updated further to be dynamic instead of static.

daralynnrhode:
# Making the output table be dynamic with sizing

## Overview

Making changes to the `_print_query_results_table()` so the sizing of the table can be dynamic based on the `query_results` received. 

Closes #53 

## Updated Files

- cli.py

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -62,39 +62,21 @@ def _print_query_results_table(query_results: list[dict]):
     ]
 
     # Calculate the maximum width for each column based on the header and the data
-    column_widths = {
-        "Instrument": max(
-            len("Instrument"),
-            *(len(str(item.get("instrument", ""))) for item in query_results),
-        ),
-        "Data Level": max(
-            len("Data Level"),
-            *(len(str(item.get("data_level", ""))) for item in query_results),
-        ),
-        "Descriptor": max(
-            len("Descriptor"),
-            *(len(str(item.get("descriptor", ""))) for item in query_results),
-        ),
-        "Start Date": max(
-            len("Start Date"),
-            *(len(str(item.get("start_date", ""))) for item in query_results),
-        ),
-        "Repointing": max(
-            len("Repointing"),
-            *(len(str(item.get("repointing", ""))) for item in query_results),
-        ),
-        "Version": max(
-            len("Version"),
-            *(len(str(item.get("version", ""))) for item in query_results),
-        ),
-        "Filename": max(
+    column_widths = {}
+    for header in headers[:-1]:
+        column_widths[header] = max(
+            len(header),
+            *(len(str(item.get(header.lower(), ""))) for item in query_results),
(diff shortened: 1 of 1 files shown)


imap_data_access/cli.py
```diff
@@ -62,31 +60,74 @@ def _print_query_results_table(query_results: list[dict]):
         "Version",
         "Filename",
     ]
-    print(format_string.format(*header))
+
+    # Calculate the maximum width for each column based on the header and the data
+    column_widths = {
+        "Instrument": max(
```
sdhoyt, line 66: Really nice way of formatting the columns widths!

sdhoyt (approved):
Looks good to me. I was tried querying with your changes and everything looked good. Great work!

imap_data_access/cli.py
```diff
@@ -62,31 +60,74 @@ def _print_query_results_table(query_results: list[dict]):
...
         ]
         print(format_string.format(*values))
+
     # Close the table
     print(hyphens)
 
 
-def extract_version_number(str_version: str) -> int:
-    """Take a version string 'vXXX' and return and int 'XXX'."""
-    return int(str_version[1:4])
```
bourque, line 87: Good catch!

imap_data_access/cli.py
```diff
@@ -62,31 +60,74 @@ def _print_query_results_table(query_results: list[dict]):
...
+            *(len(str(item.get("version", ""))) for item in query_results),
+        ),
+        "Filename": max(
+            len("Filename"),
+            *(
+                len(os.path.basename(item.get("file_path", "")))
+                for item in query_results
+            ),
+        ),
+    }
```
bourque, line 97: I was trying to think of a clever way to avoid duplicated code. Does something like this work?

```suggestion
    column_widths = {}
    for header in headers[:-1]:
        column_width[header] = max(len(header), *(len(str(item.get(header.lower(), ""))) for item in query_results)
        # Calculate the maximum width for each column based on the header and the data
     
     column_widths["Filename"] = max(len("Filename"), *(len(os.path.basename(item.get("file_path", ""))) for item in query_results)
```

bourque (approved):
Thanks for making this fix! It looks good. I just had one non-blocking suggestion if you wanted to try it out. But it is also fine as is.

imap_data_access/cli.py
daralynnrhode: It produces the same results

bourque:
@daralynnrhode It looks like this PR is ready to go, so I am going to merge this now so that it can be included in today's release.

PR #82, opened 2024-08-16
Bug Fix For CLI crash

### LINKED ISSUE #81: BUG - module not found when using CLI
### Description of the issue

There is a module-not-found error when invoking the CLI due to a relative import.

```
/lib/python3.12/site-packages/imap_data_access/cli.py", line 23, in <module>
    from file_validation import ScienceFilePath
ModuleNotFoundError: No module named 'file_validation'
```

This line needs to be prefixed with `imap_data_access.`
https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/ff6c637533d39a8c5359f2e456e918f2495b0871/imap_data_access/cli.py#L23

```
pip install imap-data-access
imap-data-access -h
```

Find all modules

crash

### Code Snippet:

_No response_

We should add a smoke test for using the CLI too. A new `test_cli.py` file? Following subprocess calls after the installation similar to https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/main/tests/test_config_options.py
We should be able to install our package and call `imap-data-access -h` without a crash.

daralynnrhode:
# Fix for the module_not_found error when installing updated Cli

## Overview

Fixing the import statement and adding a test file for the cli so the error does not slip through again.

Closes #81 
## New Files

-test_cli.py

## Updated Files

- cli.py

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -20,9 +20,8 @@ import logging
 import os
 from pathlib import Path
 
-from file_validation import ScienceFilePath
-
 import imap_data_access
+from imap_data_access.file_validation import ScienceFilePath
 
 
 def _download_parser(args: argparse.Namespace):
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,0 +1,16 @@
+"""Tests for the CLI options."""
+
+import sys
+from unittest import mock
+
+import pytest
+
+from imap_data_access import cli
+
+
+def test_cli_works():
+    """Smoke test for the CLI module making sure it is callable."""
+    with mock.patch.object(sys, "argv", ["imap-data-access", "-h"]):
+        # Should have a 0 SystemExit return code if successful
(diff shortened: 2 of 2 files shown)


greglucas:
Could you add this in a `test_cli.py` file? This fails on main and passes on this. Something like this is easy to slip through the cracks, so nice to have a quick smoke test to verify we don't accidentally break it again in the future.

```python
"""Tests for the CLI options."""

import sys
from unittest import mock

import pytest

from imap_data_access import cli


def test_cli_works():
    """Smoke test for the CLI module making sure it is callable."""
    with mock.patch.object(sys, "argv", ["imap-data-access", "-h"]):
        # Should have a 0 SystemExit return code if successful
        with pytest.raises(SystemExit, match="0"):
            cli.main()
```

bourque (approved):
Thanks for fixing this!

greglucas (approved):
Awesome, thanks for the very quick fix!

PR #84, opened 2024-08-19
REL: v0.8.1

bourque:
Updated version numbers for release v0.8.1


greglucas:
I wasn't entirely sure which version number we wanted to bump, but I _think_ we should do `0.9.0` since there are new "features" here as well and it isn't just the CLI fix.

PR #86, opened 2024-08-31
[WIP] Adding a Pandas.DataFrame output option for Python Query

### LINKED ISSUE #73: Allow option for table output when using python package to query
Currently the `query()` method in the python tool returns a list of dictionaries containing the results of the query. It would be nice to allow a user to pass some option (e.g. `output='table'` or `output='df'`) that will return a pandas dataframe instead. Some users may prefer to work with a dataframe instead of a list of dicts.

daralynnrhode:
# Giving the option to change the output of a python query to be a pandas.DataFrame

## Overview

I have not been able to test this myself using the python import, but wanted to get the PR up incase I missed anything.

Added a new parameter to the `query()` in `io.py` to dictate what the return type for the function should be. If passed in 'df' then the `list[dict[]]` will be converted to a dataFrame before being returned.  

Closes #73 
## Updated Files

- io.py

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -12,6 +12,8 @@ from typing import Optional, Union
 from urllib.error import HTTPError, URLError
 from urllib.parse import urlencode
 
+import pandas as pd
+
 import imap_data_access
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


greglucas:
Sorry I haven't kept up with the tickets, so didn't comment on the original issue. But, I dislike this update here. I think this package should be kept extremely minimal with no dependencies if we can help it.

My suggestion would be to add documentation for how to fill your pandas dataframe yourself, which is actually easier to remember than having to specify a new keyword argument IMO.

```python
# The query response can be turned into a pandas DataFrame directly for ease of future use too
df = pd.DataFrame(imap_data_access.query(instrument="swapi")
```

PR #87, opened 2024-09-07
Input validation to handle 502 Errors

### LINKED ISSUE #55: Data Access API 502 errors
Provide more informative error messages for when this error is thrown in certain cases. See User Testing #40 for more information and clearer instances when it occurs.

daralynnrhode:
# Adding input validation for some parameters to avoid 502 errors

## Overview

Input validation code for all input parameters. This should avoid most if not all 502 errors triggered when incorrect data is passed to the query and it crashes. 
Also introduced a test to iterate over bad input values and check that the correct valueError message is returned. 

Closes #55 
## Updated Files

- io.cli
- tests/test_io.py

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -159,6 +159,18 @@ def query(
             "At least one query parameter must be provided. "
             "Run 'query -h' for more information."
         )
+    print(query_params)
+    # print(query_params["data_level"])
+    # print(query_params["data_level"][0])
+    # Check data-level
+    # do an if statement that checks that data_level was passed in,
+    # then check it against all options, l0, l1a, l1b, l2, l3 etc.
+
+    # Check start-date
+    # Check end-date
+    # Check version
+    # Check filename
+
     url = f"{imap_data_access.config['DATA_ACCESS_URL']}"
     url += f"/query?{urlencode(query_params)}"
 
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -186,7 +186,8 @@ def query(
 
     # Check version
     if version is not None:
-        raise ValueError("Not a valid version, use format 'vXXX'.")
+        if version[0] != "v" and len(version) != 4:
+            raise ValueError("Not a valid version, use format 'vXXX'.")
 
     url = f"{imap_data_access.config['DATA_ACCESS_URL']}"
     url += f"/query?{urlencode(query_params)}"
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -202,7 +202,7 @@ def query(
         and not file_validation.ScienceFilePath.is_valid_repointing(repointing)
     ):
         raise ValueError(
-            "Not a valid repointing, use format repointing<num>,"
+            "Not a valid repointing, use format repoint<num>,"
             " where <num> is a 5 digit integer."
         )
 
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -159,20 +159,25 @@ def query(
             "At least one query parameter must be provided. "
             "Run 'query -h' for more information."
         )
-    print(query_params)
-    # print(query_params["data_level"])
-    # print(query_params["data_level"][0])
     # Check data-level
     # do an if statement that checks that data_level was passed in,
-    if data_level is not None:
-        if data_level not in ("l0", "l1a", "l1b", "l2", ""):
-            raise ValueError("Not a valid data level, choose from .......")
     # then check it against all options, l0, l1a, l1b, l2, l3 etc.
+    if data_level is not None:
+        if data_level not in ("l0", "l1a", "l1b", "l2", "l3"):
+            raise ValueError(
+                "Not a valid data level, choose from 'l0', 'l1a', 'l1b', 'l2', 'l3'."
+            )
+    # will need all options for data-levels
 
     # Check start-date
+    if start_date is not None:
+        raise ValueError("Not a valid start date, use format 'YYYYMMDD'.")
     # Check end-date
+    if end_date is not None:
+        raise ValueError("Not a valid end date, use format 'YYYYMMDD'.")
     # Check version
-    # Check filename
+    if version is not None:
+        raise ValueError("Not a valid version, use format 'vXXX'.")
 
     url = f"{imap_data_access.config['DATA_ACCESS_URL']}"
     url += f"/query?{urlencode(query_params)}"
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -184,9 +184,9 @@ def query(
         except ValueError as err:
             raise ValueError("Not a valid end date, use format 'YYYYMMDD'.") from err
 
-    # Check version
+    # Check version make sure to include 'latest'
     if version is not None:
-        if version[0] != "v" and len(version) != 4:
+        if version != "latest" and (version[0] != "v" and len(version) != 4):
             raise ValueError("Not a valid version, use format 'vXXX'.")
 
     url = f"{imap_data_access.config['DATA_ACCESS_URL']}"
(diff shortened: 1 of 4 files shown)


daralynnrhode:
I went in and made a small change to the `test_io.py``query_parameters` supplied to the tests. The `version` being supplied was `000` which does not adhere to the code I introduced. The new code requires `version` be passed in as `v000`. I opted to change the test to reflect the formatting that was consistently being used, but I can aways reverse the change and update the code as well as any documentation. Although, this would require both documentation and output messages to be updated as well.

imap_data_access/io.py
```diff
@@ -159,6 +163,85 @@ def query(
             "At least one query parameter must be provided. "
             "Run 'query -h' for more information."
         )
+    # Check instrument name
+    if instrument is not None:
+        if instrument not in (
```
maxinelasp, line 168: Can you get this from `VALID_INSTRUMENTS` https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/main/imap_data_access/__init__.py

imap_data_access/io.py
```diff
@@ -159,6 +163,85 @@ def query(
...
+            "idex",
+            "lo",
+            "mag",
+            "swapi",
+            "swe",
+            "ultra",
+        ):
+            raise ValueError(
+                "Not a valid instrument, please choose from "
+                "('codice', 'glows', 'hi', 'hit', 'idex', "
```
maxinelasp, line 182: Similarly, just print out the valid instruments here. Not that I think they're going to start bolting instruments on, but we have had this list change before

imap_data_access/io.py
```diff
@@ -159,6 +163,85 @@ def query(
...
+                "'lo', 'mag', 'swapi', 'swe', 'ultra')"
+            )
+
+    # Check data-level
+    # do an if statement that checks that data_level was passed in,
+    # then check it against all options, l0, l1a, l1b, l2, l3 etc.
+    if data_level is not None:
+        if data_level not in (
```
maxinelasp, line 190: Ditto for valid data levels. I actually need to remove L2pre, so it would be good to put it in one place

imap_data_access/io.py
```diff
@@ -159,6 +163,85 @@ def query(
...
+            "l3d",
+        ):
+            raise ValueError(
+                "Not a valid data level, choose from "
+                "('l0','l1','l1a','l1b','l1c','l1ca','l1cb',"
+                "'l1d','l2','l2pre','l3','l3a','l3b','l3c','l3d')."
+            )
+
+    # Check start-date
+    if start_date is not None:
```
maxinelasp, line 214: For dates, can you use `is_valid_date` from ScienceFilePath? https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/main/imap_data_access/file_validation.py

imap_data_access/io.py
```diff
@@ -159,6 +163,85 @@ def query(
...
+
+    # Check end-date
+    if end_date is not None:
+        try:
+            datetime.strptime(end_date, "%Y%m%d")
+        except ValueError as err:
+            raise ValueError("Not a valid end date, use format 'YYYYMMDD'.") from err
+
+    # Check version make sure to include 'latest'
+    if version is not None:
```
maxinelasp, line 228: Similarly, could you put this check in some kind of `is_valid_version` in ScienceFilePath so we can share it around?

imap_data_access/io.py
```diff
@@ -159,6 +163,85 @@ def query(
...
+        if version != "latest" and (version[0] != "v" and len(version) != 4):
+            raise ValueError("Not a valid version, use format 'vXXX'.")
+
+    # check repointing follows 'repoint00000' format
+    if repointing is not None:
```
maxinelasp, line 233: Ditto for some kind of `is_valid_repointing`. This would be useful for the filename checks.

imap_data_access/io.py
```diff
@@ -159,6 +163,85 @@ def query(
...
+        if not re.fullmatch(r"repoint\d{5}", str(repointing)):
+            raise ValueError(
+                "Not a valid repointing, use format repointing<num>,"
+                " where <num> is a 5 digit integer."
+            )
+
+    # check extension
+    if extension is not None:
```
maxinelasp, line 241: this is also in `__init__` as `VALID_FILE_EXTENSIONS`

tests/test_io.py
```diff
@@ -218,6 +219,67 @@ def test_query_bad_params(mock_urlopen: unittest.mock.MagicMock):
...
+        ),
+        ("version", "badInput", "Not a valid version, use format 'vXXX'."),
+        (
+            "extension",
+            "badInput",
+            re.escape("Not a valid extension, choose from ('pkts', 'cdf')."),
+        ),
+    ],
+)
+def test_bad_query_input(query_flag, query_input, expected_output):
```
maxinelasp, line 260: Nice parameter test :)

maxinelasp (changes requested):
Had a few comments, mostly moving to some shared pieces that makes your code cleaner :) Hopefully this can remove that ruff ignore statement too, no big deal if it doesn't though

maxinelasp (approved):
There is one comment that needs updating but other than that, looks good! Thanks for making those changes!

tests/test_io.py
```diff
@@ -218,6 +219,61 @@ def test_query_bad_params(mock_urlopen: unittest.mock.MagicMock):
...
+    expected_output : str
+        Output error expected to be given.
+    """
+    kwargs = {query_flag: query_input}
+
+    # Check if the ValueError is raised and contains the correct message
+    with pytest.raises(ValueError, match=expected_output):
+        imap_data_access.query(**kwargs)
+
+    # assert str(excinfo.value) == expected_output
```
bourque, line 274: I think you can remove this if it is not being used

bourque (approved):
Awesome work here, this looks great! And great work on making that parameterized test. That is some advanced level unit testing 🔥

tests/test_io.py
```diff
@@ -218,6 +219,61 @@ def test_query_bad_params(mock_urlopen: unittest.mock.MagicMock):
...
 
+@pytest.mark.parametrize(
+    ("query_flag", "query_input", "expected_output"),
+    [
+        # All parameters should  not send query
+        (
+            "instrument",
+            "badInput",
+            "Not a valid instrument, please choose from "
+            + ", ".join(imap_data_access.VALID_INSTRUMENTS),
```
greglucas, line 230: No need to update here, but as an FYI in case it simplifies things later for you. `pytest.raises(match=x)` will do partial matches for you. So often I'll just put the first few words down that are what matters rather than the exact string which can get cumbersome.

```suggestion
            "Not a valid instrument"
```
and your ones later would be like "not a valid start date" because the start / end / instrument are meaningful to your message, but the exact instruments not so much.

greglucas (commented):
Looks great, just some very minor stylistic comments that you can take or leave.

imap_data_access/io.py
```diff
@@ -159,6 +162,54 @@ def query(
...
+    ):
+        raise ValueError("Not a valid version, use format 'vXXX'.")
+
+    # check repointing follows 'repoint00000' format
+    if (
+        repointing is not None
+        and not file_validation.ScienceFilePath.is_valid_repointing(repointing)
+    ):
+        raise ValueError(
+            "Not a valid repointing, use format repointing<num>,"
```
greglucas, line 205: Remove the `ing` and make it explicit how many numbers there has to be. (Update the actual code as well, I just noticed this in the test here)
```suggestion
            "Not a valid repointing, use format repointXXXXX,"
```

greglucas (approved):
Thanks for making those minor updates, I think it reads nicely now. I found one minor thing in the error string example that should be updated too, otherwise looks great!

PR #88, opened 2024-09-13
MNT: Follow 307 redirects from AWS

greglucas:
When deploying a new bucket in a non us-east-1 region, it can take time to propagate to propagate the DNS records and thus AWS will give us a 307 redirect to a region-specific endpoint. This only matters for PUT requests because those aren't supposed to be followed by default and therefore Python raises an HTTPError with a 307 code that we catch. So we can do the redirect ourselves from within there.

The attempted fix in `sds-data-manager` didn't fix the issue apparently when deploying fresh this morning. This does fix the issue from within `imap-data-access` instead.

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -37,11 +37,18 @@ def _get_url_response(request: urllib.request.Request):
             yield response
 
     except HTTPError as e:
-        message = (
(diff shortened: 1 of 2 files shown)


tech3371 (approved):
This looks good! Sorry for late review. I looked at it but got distracted with other things and forgot.

PR #89, opened 2024-09-20
MNT/FIX: Removing imap/ prefix from SPICE files

greglucas:
It is currently `/imap/spice/ck/...`
We are changing it to `/spice/ck/...`
and dropping the imap/ prefix to separate out spice into its own separate directory.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -354,5 +354,5 @@ class SPICEFilePath:
         spice_dir = imap_data_access.config["DATA_DIR"] / "spice"
         subdir = _SPICE_DIR_MAPPING[self.filename.suffix]
         # Use the file suffix to determine the directory structure
-        # IMAP_DATA_DIR/imap/spice/<subdir>/filename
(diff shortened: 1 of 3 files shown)


imap_data_access/file_validation.py
```diff
@@ -351,7 +351,7 @@ def construct_path(self) -> Path:
         Path
             Upload path
         """
-        spice_dir = imap_data_access.config["DATA_DIR"] / "imap/spice"
+        spice_dir = imap_data_access.config["DATA_DIR"] / "spice"
         subdir = _SPICE_DIR_MAPPING[self.filename.suffix]
         # Use the file suffix to determine the directory structure
         # IMAP_DATA_DIR/imap/spice/<subdir>/filename
```
tech3371, line 357:
```suggestion
        # IMAP_DATA_DIR/spice/<subdir>/filename
```

tech3371 (approved):
It looks good. If we go this path, I opened new PR that matches this update here, https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/345

PR #94, opened 2024-10-15
Support for Ancillary Files, Creating Ancillary File Class

daralynnrhode:
# Adding an Ancillary File class to file validation

## Overview

Using the ScienceFile class as a model I will be implementing similar code to support the upload and downloads of ancillary files with the api. 

## Updated Files

- __init__.py
- file_validation.py
- io.py

### DIFF
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -81,3 +81,8 @@ FILENAME_CONVENTION = (
     "<mission>_<instrument>_<datalevel>_<descriptor>_"
     "<startdate>(-<repointing>)_<version>.<extension>"
 )
+
+ANCILLARY_FILENAME_CONVENTION = (
+    "<mission>_<instrument>_<ancillary_name>_"
+    "<start_date>_<end_date>_<version>.<extension>"
+)
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -84,5 +84,5 @@ FILENAME_CONVENTION = (
 
 ANCILLARY_FILENAME_CONVENTION = (
     "<mission>_<instrument>_<ancillary_name>_"
-    "<start_date>_<end_date>_<version>.<extension>"
+    "<start_date>(-<end_date>)_<version>.<extension>"
 )
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -85,6 +85,8 @@ FILENAME_CONVENTION = (
 )
 
 ANCILLARY_FILENAME_CONVENTION = (
-    "<mission>_<instrument>_<ancillary_name>_"
+    "<mission>_<instrument>_<description>_"
     "<start_date>(-<end_date>)_<version>.<extension>"
 )
+
+VALID_ANCILLARY_FILE_EXTENSION = {"cdf", "csv", "json"}
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -356,3 +356,273 @@ class SPICEFilePath:
         # Use the file suffix to determine the directory structure
         # IMAP_DATA_DIR/imap/spice/<subdir>/filename
         return spice_dir / subdir / self.filename
+
+
+class AncillaryFilePath:
+    """Class for building and validating filepaths for Ancillary files."""
+
+    class InvalidAncillaryFileError(Exception):
+        """Indicates a bad file type."""
+
+        pass
+
+    def __init__(self, filename: str | Path):
+        """Class to store filepath and file management methods for Ancillary files.
+
+        If you have an instance of this class, you can be confident you have a valid
+        ancillary file and generate paths in the correct format. The parent of the file
+        path is set by the "IMAP_DATA_DIR" environment variable, or defaults to "data/"
+
+        Current filename convention:
+        "<mission>_<instrument>_<ancillary_name>_<start_date>(-<end_date>)_
+        <version>.<extension>"
+
+        <mission>: imap
+        <instrument>: codice, glows, hi, hit, idex, lo, mag, swapi, swe, ultra
+        <ancillary_name>: A descriptive name for the ancillary file which
+        distinguishes between other ancillary files used by the instrument.
+        <start_date>: startdate is the earliest date in the data, format: YYYYMMDD
+        <end_date>: The end time of the validity of the ancillary file,
+        in the format “YYYYMMDD”. This is optional for files, with the understanding
+         that if end_date is not provided, the file is valid until a file with a
+         later start_date and no end_date.
+        <version>: This stores the data version for this product, format: vXXX
+
+        Parameters
+        ----------
+        filename : str | Path
+            Ancillary data filename or file path.
+        """
+        self.filename = Path(filename)
+        self.data_dir = imap_data_access.config["DATA_DIR"]
+
+        try:
+            split_filename = self.extract_filename_components(self.filename)
+        except ValueError as err:
+            raise self.InvalidAncillaryFileError(
+                f"Invalid filename. Expected file to match format: "
+                f"{imap_data_access.ANCILLARY_FILENAME_CONVENTION}"
+            ) from err
+
+        self.mission = split_filename["mission"]
+        self.instrument = split_filename["instrument"]
+        self.ancillary_name = split_filename["ancillary_name"]
+        self.start_date = split_filename["start_date"]
+        self.end_date = split_filename["end_date"]
+        self.version = split_filename["version"]
+        self.extension = split_filename["extension"]
+
+        self.error_message = self.validate_filename()
+        if self.error_message:
+            raise self.InvalidAncillaryFileError(f"{self.error_message}")
+
+    """Changes needed below here."""
+
+    @classmethod
+    def generate_from_inputs(
+        cls,
+        instrument: str,
+        data_level: str,
+        descriptor: str,
+        start_time: str,
+        version: str,
+        repointing: int | None = None,
+    ) -> AncillaryFilePath:
+        """Generate filename from given inputs and return a AncillaryFilePath instance.
+
+        This can be used instead of the __init__ method to make a new instance:
+        ```
+        ancillary_file_path = AncillaryFilePath.generate_from_inputs("mag", "l0",
+            "test", "20240213", "v001")
+        full_path = ancillary_file_path.construct_path()
+        ```
+        *CHANGE BELOW
+
+        Parameters
+        ----------
+        descriptor : str
+            The descriptor for the filename
+        instrument : str
+            The instrument for the filename
+        data_level : str
+            The data level for the filename
+        start_time: str
+            The start time for the filename
+        version : str
+            The version of the data
+        repointing : int, optional
+            The repointing number for this file, optional field that
+            is not always present
+
+        Returns
+        -------
+        str
+            The generated filename
+        """
+        extension = "cdf"
+        if data_level == "l0":
+            extension = "pkts"
+        time_field = start_time
+        if repointing:
+            time_field += f"-repoint{repointing:05d}"
+        filename = (
+            f"imap_{instrument}_{data_level}_{descriptor}_{time_field}_"
+            f"{version}.{extension}"
+        )
+        return cls(filename)
+
+    def validate_filename(self) -> str:
+        """Validate the filename and populate the error message for wrong attributes.
+
+        The error message will be an empty string if the filename is valid. Otherwise,
+        all errors with the filename will be put into the error message.
+
+        Returns
+        -------
+        error_message: str
+            Error message for specific missing attribute, or "" if the file name is
+            valid.
+        """
+        error_message = ""
+
+        if any(
+            attr is None or attr == ""
+            for attr in [
+                self.mission,
+                self.instrument,
+                self.data_level,
+                self.descriptor,
+                self.start_date,
+                self.version,
+                self.extension,
+            ]
+        ):
+            error_message = (
+                f"Invalid filename, missing attribute. Filename "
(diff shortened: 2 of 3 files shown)


imap_data_access/file_validation.py
```diff
@@ -538,7 +545,9 @@ def is_valid_date(input_date: str) -> bool:
 
     """Changes needed below here."""
 
-    def construct_path(self) -> Path:
+    def construct_path(
+        self,
+    ) -> Path:  # TODO: how should this path look? Should it be different format?
```
maxinelasp, line 550: Good question - this should be a different function or maybe a flag to differentiate ancillary files and data files. I'm going to ask in slack about where ancillary files should live. 

The initial vote seems to be `imap/ancillary/[instrument]` which also seems like the most logical spot for me. Keep an eye on slack in case that changes, but that's a fine assumption for now.

imap_data_access/__init__.py
```diff
@@ -81,3 +81,8 @@
     "<mission>_<instrument>_<datalevel>_<descriptor>_"
     "<startdate>(-<repointing>)_<version>.<extension>"
 )
+
+ANCILLARY_FILENAME_CONVENTION = (
+    "<mission>_<instrument>_<ancillary_name>_"
```
maxinelasp, line 86: Nice, minor nitpick would be to replace "ancillary_name" with "descriptor" just to make it a little more clear.

imap_data_access/file_validation.py
```diff
@@ -356,3 +356,259 @@ def construct_path(self) -> Path:
         # Use the file suffix to determine the directory structure
         # IMAP_DATA_DIR/imap/spice/<subdir>/filename
         return spice_dir / subdir / self.filename
+
+
+class AncillaryFilePath:
```
maxinelasp, line 361: Nice, I like the new class. Definitely should inherit from ScienceFilePath and overwrite functionality for two reasons: 
1. an ancillary file is a type of science file, and has very similar functionality, so having it inherit would make that clear since the data model is that an ancillary file is a subset of a science file. Ask me about this in our next student meeting so I can remember to talk about this a little more - it's important for design work like this
2. you can share some of the functionality of the original class - for example, the validation stuff

imap_data_access/file_validation.py
```diff
@@ -356,3 +356,259 @@ def construct_path(self) -> Path:
...
+            split_filename = self.extract_filename_components(self.filename)
+        except ValueError as err:
+            raise self.InvalidAncillaryFileError(
+                f"Invalid filename. Expected file to match format: "
+                f"{imap_data_access.ANCILLARY_FILENAME_CONVENTION}"
+            ) from err
+
+        self.mission = split_filename["mission"]
+        self.instrument = split_filename["instrument"]
+        self.ancillary_name = split_filename["ancillary_name"]
```
maxinelasp, line 409: If we make this 'descriptor" then we can actually match all the attributes of ScienceFilePath and share a lot more code. Plus, this piece does still act as a descriptor.

imap_data_access/file_validation.py
```diff
@@ -356,3 +356,259 @@ def construct_path(self) -> Path:
...
+
+        if not self.is_valid_date(self.start_date):
+            error_message += "Invalid start date format. Please use YYYYMMDD format. \n"
+        if not self.is_valid_date(self.end_date):
+            error_message += "Invalid end date format. Please use YYYYMMDD format. \n"
+
+        return error_message
+
+    @staticmethod
+    def is_valid_date(input_date: str) -> bool:
```
maxinelasp, line 524: Should use the version from ScienceFilePath rather than duplicating it. Since that method is static, you can replace any calls to `self.is_valid_date()` with `ScienceFilePath.is_valid_date()` and it will use the already-existing method (that is why it is static in ScienceFilePath.)

If you want to use a non-static method, you can inherit the class and methods and then use it that way. Have you taken a class in object oriented coding? I am happy to hop on a call and explain some of the basics.

maxinelasp (commented):
Good start! I had a few comments and suggestions, let me know if you would like to meet to discuss the design further.

imap_data_access/__init__.py
```diff
@@ -83,3 +83,10 @@
     "<mission>_<instrument>_<datalevel>_<descriptor>_"
     "<startdate>(-<repointing>)_<version>.<extension>"
 )
+
+ANCILLARY_FILENAME_CONVENTION = (
+    "<mission>_<instrument>_<description>_"
+    "<start_date>(-<end_date>)_<version>.<extension>"
+)
+
+VALID_ANCILLARY_FILE_EXTENSION = {"cdf", "csv", "json"}
```
bourque, line 92:
```suggestion
VALID_ANCILLARY_FILE_EXTENSIONS = {"cdf", "csv", "json"}
```

imap_data_access/io.py
```diff
@@ -75,8 +75,26 @@ def download(file_path: Union[Path, str]) -> Path:
...
+                # Matches neither file format
+                logger.error(
+                    "Invalid file type: %s. File does not match Science or "
+                    "Ancillary file formats.",
+                    file_path,
+                )
+                raise ValueError(
+                    f"Invalid file: {file_path}. It does not match Science or "
+                    f"Ancillary file formats."
+                ) from e2
```
bourque, line 97: These look to be basically the same error message? So I think you could do something like this to make it more concise.

```suggestion
            except imap_data_access.AncillaryFilePath.InvalidAncillaryFileError as error:
                # Matches neither file format
                error_message = f"Invalid file type for {file_path}. It does not match Science or Ancillary file formats"
                logger.error(error_message)
                raise ValueError(error_message) from e2
```

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+        <mission>: imap
+        <instrument>: codice, glows, hi, hit, idex, lo, mag, swapi, swe, ultra
+        <description>: A descriptive name for the ancillary file which
+        distinguishes between other ancillary files used by the instrument.
+        <start_date>: startdate is the earliest date in the data, format: YYYYMMDD
+        <end_date>: The end time of the validity of the ancillary file,
+        in the format “YYYYMMDD”. This is optional for files, with the understanding
+         that if end_date is not provided, the file is valid until a file with a
+         later start_date and no end_date.
+        <version>: This stores the data version for this product, format: vXXX
```
bourque, line 430:
```suggestion
        <mission>: imap
        <instrument>: codice, glows, hi, hit, idex, lo, mag, swapi, swe, ultra
        <description>: A descriptive name for the ancillary file which
                       distinguishes between other ancillary files used by the
                       instrument.
        <start_date>: startdate is the earliest date in the data, format: YYYYMMDD
        <end_date>: The end time of the validity of the ancillary file,
                    in the format “YYYYMMDD”. This is optional for files, with the
                    understanding that if end_date is not provided, the file is valid
                    until a file with a later start_date and no end_date.
        <version>: This stores the data version for this product, format: vXXX
```

Small indentation suggestion for improved readability

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+            The start time for the filename. If not provided
+            it is assumed to be valid for all times.
+        end_time: str, optional
+            The end time for the filename. If not provided,
+            the file is valid until a file with a later
+            start_date and no end_date.
+        version : str
+            The version of the data.
+        extension : str
+            The extension type of the file.
```
bourque, line 495: It would be good to list these in the order that they are given in the method definition: instrument, description, version, extension, start_time, end_time. Or change the order of them in the method definition to match these, either way.

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+        -------
+        str
+            The generated filename
+        """
+        time_field = start_time
+        if end_time:
+            time_field += end_time  # I think...
+        filename = (
+            f"imap_{instrument}_{description}_{time_field}_" f"{version}.{extension}"
+        )
```
bourque, line 507: Isn't the `end_time` just a `YYYYMMDD` formatted date (or `None`) provided by the user? So I think you could just do:

```suggestion
        if end_time:
            filename = (
                f"imap_{instrument}_{description}_{start_time}-{end_time}_{version}.{extension}"
            )
        else:
           filename = (
                f"imap_{instrument}_{description}_{start_time}_{version}.{extension}"
            )
```

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+                self.instrument,
+                self.description,
+                self.version,
+                self.extension,
+            ]
+        ):
+            error_message = (
+                f"Invalid filename, missing attribute. Filename "
+                f"convention is {imap_data_access.FILENAME_CONVENTION} \n"
+            )
```
bourque, line 537: This is clever!

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+        if self.mission != "imap":
+            error_message += f"Invalid mission {self.mission}. Please use imap \n"
+
+        if self.instrument not in imap_data_access.VALID_INSTRUMENTS:
+            error_message += (
+                f"Invalid instrument {self.instrument}. Please choose "
+                f"from "
```
bourque, line 544:
```suggestion
                f"Invalid instrument {self.instrument}. Please choose from "
```

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+                f"{imap_data_access.VALID_INSTRUMENTS} \n"
+            )
+
+        if self.extension not in imap_data_access.VALID_FILE_EXTENSION:
+            error_message += (
+                "Invalid extension. Extension should be cdf. \n"  # TODO: Change this
```
bourque, line 550: Yeah you should be able to utilize the `VALID_ANCILLARY_FILE_EXTENSION` variable here

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+            upload_path = self.data_dir / upload_path
+
+        return upload_path
+
+    @staticmethod
+    def extract_filename_components(filename: str | Path) -> dict:
+        """Extract all components from filename. Does not validate instrument or level.
+
+        Will return a dictionary with the following keys:
+        { instrument, ancillary_name, startdate, enddate, version, extension }
```
bourque, line 586:
```suggestion
        { instrument, description, start_date, end_date, version, extension }
```

bourque (commented):
Awesome work here! This is tricky stuff and you nailed it.

I have some formatting suggestions and nitpicks, but nothing major.

imap_data_access/__init__.py
maxinelasp: I would actually say we should leave it `EXTENSION` to match the existing typo `VALID_FILE_EXTENSION` 😅 but maybe that's crazy

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+
+        Current filename convention:
+        "<mission>_<instrument>_<description>_<start_date>(-<end_date>)_
+        <version>.<extension>"
+
+        <mission>: imap
+        <instrument>: codice, glows, hi, hit, idex, lo, mag, swapi, swe, ultra
+        <description>: A descriptive name for the ancillary file which
+        distinguishes between other ancillary files used by the instrument.
+        <start_date>: startdate is the earliest date in the data, format: YYYYMMDD
```
maxinelasp, line 425:
```suggestion
        <start_date>: startdate is the earliest date where the file is valid, format: YYYYMMDD
```
nitpick

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+        <end_date>: The end time of the validity of the ancillary file,
+        in the format “YYYYMMDD”. This is optional for files, with the understanding
+         that if end_date is not provided, the file is valid until a file with a
+         later start_date and no end_date.
```
maxinelasp, line 429: <3 this comment, super clear

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+        self.start_date = split_filename["start_date"]
+        self.end_date = split_filename["end_date"]
+        self.version = split_filename["version"]
+        self.extension = split_filename["extension"]
+
+        self.error_message = self.validate_filename()
+        if self.error_message:
+            raise self.InvalidAncillaryFileError(f"{self.error_message}")
+
+    @classmethod
```
maxinelasp, line 460: Let me know if you need me to explain class methods vs static methods and why to use one or the other... 😅 I forgot that I used one in the ScienceFilePath class. This style is called a "factory pattern" if you wanna Google more about it!

imap_data_access/file_validation.py
```diff
@@ -397,3 +397,227 @@ def construct_path(self) -> Path:
...
+        full_path = ancillary_file_path.construct_path()
+        ```
+
+        Parameters
+        ----------
+        description : str
+            The descriptor for the ancillary filename.
+        instrument : str
+            The instrument for the filename.
+        start_time: str, optional
```
maxinelasp, line 485: You have start time optional here, but required in the method signature. It should be required, I think... we could have it default to launch date, but that requires extra work so I don't think it's needed yet.

maxinelasp (commented):
Looks really really good! I agree with Matthew's comments and made a few minor ones, but nice work!

maxinelasp (approved):
Great job!

bourque (approved):
Nice work, looks great!

Issue #109, opened 2025-01-22
Write an abstraction layer for general dependencies and ancillary files


maxinelasp:
### Summary

It would be convienent to have an abstraction layer which describes a dependency without specifying if it is a science, ancillary, or SPICE file. 

This dependency super class should have methods/variables for:
- valid date range
- instrument (or all)
- filenames 

Then, we should also begin work on a shared ancillary file dependency manager class to create a place to handle the complex combinations of multiple ancillary files into one piece of data. 

This ticket should only cover the very general case. It is intended to act as a general class which we can then develop against for more specific use cases as we re-work dependency manager and IMAP CLI. 

This is in IMAP data access because it is a framework for accessing data files, and I anticipate it will be used as part of the IMAP CLI and as batch-starter. 

### Requirements

The more specific the better.
  - [ ] Create a general dependencies class
  - [ ] Maybe come up with a better name for it
  - [ ] Write a basic ancillary file manager that inherits from it
  - [ ] Sketch out the basics of SPICE and science file classes to have a place to put stuff

tech3371:
I want to add note about shared reading tools idea we discussed. I don't know if that fits here.

Some cases we discussed:
1. If cdf, load and return xr.Dataset
2. If csv, load and return pd.DataFrame
3. If spice, load and return kernel pool?
4. others that I don't about

tech3371:
@lacoak21 This will be helpful

tech3371:
This has been completed by Input collection class feature in `imap-data-access`.

PR #115, opened 2025-02-04
First pass at shared dependency management interface

maxinelasp:
# Change Summary
As Tenzin and I were discussing the dependency system in AWS, I was thinking it might be good to have a tool which abstracts away the type of dependency file when we're passing them through batch_starter and into the CLI. This tool would collect the files into logical groups and provide a common interface when dealing with the three different types of files, so ancillary, science, and spice files would all look the same from the outside. This would be particularly nice for ancillary files as we will need to have multiple ancillary files act as one input.

This is the first draft of what that might look like so we can decide if it's worth pursuing further.

This is now completed and is ready for use and future extension. 

## Overview

processing_input_file.py contains the new classes (will be used in both imap_processing and sds-data-manager.) temp_example.py illustrates what usage would look like. 

Partly/mostly completes #109


imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,195 @@
...
+            if file_creator["type"] == ProcessingInputType.SCIENCE_FILE.value:
+                print(type(file_creator["path"]))
+                print(type(file_creator["path"][0]))
+                self.files.append(ScienceInput(*file_creator["path"]))
+            elif file_creator["type"] == ProcessingInputType.ANCILLARY_FILE.value:
+                self.files.append(AncillaryInput(*file_creator["path"]))
+            elif file_creator["type"] == ProcessingInputType.SPICE_FILE.value:
+                self.files.append(SpiceInput(*file_creator["path"]))
+
+
```
tech3371, line 195: I will try to note some of discussion points I have regarding this.
1. I am thinking from both batch starter and cli point. Can we make template class to take input such as these
```
source: str
data_type: str
descriptor: str
start_date: str
end_date: str
```
Then make query through query API and return files list. Right now, `ScienceInput` expects files list, but it would be great for both batch starter and cli to get back files list from this parent function. This will allow both to behave same and also reduce arguments passed to `--dependency` argument of cli. Batch job has character limitation on command passed in. I see this becoming issue in reprocessing or some other scenarios that I don't yet.
2.

greglucas:
Is this related to #109 ?

imap_data_access/processing_input_file.py
tech3371: What is `cli.ProcessInstrument.pre_processing()` returning now with this change?
Instrument team may need to overwrite `pre_processing()` for specific pieces. What would be useful to return and what would be share? This is question we should think about when thinking about above question.

`pre_processing()` could be shared.

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,348 @@
...
+        str
+            A string of JSON-formatted serialized output.
+        """
+        json_out = []
+        for file in self.processing_input:
+            json_out.append(file.construct_json_output())
+
+        return json.dumps(json_out)
+
+    def deserialize(self, json_input: str) -> None:
```
lacoak21, line 314: When would deserialize be used?

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,348 @@
...
+        ----------
+        processing_inputs : list | ProcessingInput
+            Either a list of ProcessingInputs or a single ProcessingInput instance.
+        """
+        if isinstance(processing_inputs, list):
+            self.processing_input.extend(processing_inputs)
+        else:
+            self.processing_input.append(processing_inputs)
+
+    def serialize(self) -> str:
```
lacoak21, line 299: I can see this being super useful for the dependency api. Maybe we could serialize the inputs and use those directly as the request query parameters! ( As long data_type, source, and descriptor are there.)

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,348 @@
...
+        list[str]
+            List of filenames that are required for the given day.
+        """
+        # todo: complete this
+        return NotImplementedError
+
+
+class SpiceInput(ProcessingInput):
+    """
+    SPICE file subclass for ProcessingInput.
```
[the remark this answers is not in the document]
maxinelasp: Because I forgot 😅 the SPICE work is the least complete because SPICEFilePath is significantly less featured than the ones for ancillary and science files, and I don't know the complete requiremetns for SPICE files. But I will add the input type here!

imap_data_access/processing_input_file.py
maxinelasp: The basic steps would be:
1. batch_starter assembles all the files and creates a ProcessingInputCollection object to pass along
2. batch_starter calls ProcessingInputCollection.serialize() and then passes the output of serialize() into the imap_processing CLI
3. imap_processing takes that output and passes it into deserialize()
4. imap_processing now has the same ProcessingInputCollection object as from batch starter and can do things like validate the required time range, check that all files that are needed for that stage in processing are present, and download all the required files.

maxinelasp: So this is basically to convert the entire object into a short string that can be passed around as necessary within AWS.

imap_data_access/processing_input_file.py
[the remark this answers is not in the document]
lacoak21: AH got it!

tech3371 (changes requested):

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,387 @@
...
+    # Following three are retrieved from dependency check.
+    # But they can also come from the filename.
+    source: str = field(init=False)
+    data_type: str = field(init=False)  # should be data level or "ancillary" or "spice"
+    descriptor: str = field(init=False)
+
+    def __init__(self, *args):
+        """Initialize using a list of filepaths and sets the attributes of the class.
+
+        This method works for ScienceFilePaths and AncillaryFilePaths. Subclasses
```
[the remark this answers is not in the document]
maxinelasp: I think SPICE will need to do something different since the filename is different. This function can be overridden in SPICEInput to handle those changes.

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,387 @@
...
+        """
+        # For science and ancillary files
+        source = set()
+        data_type = set()
+        descriptor = set()
+        file_path_list = []
+        for file in self.filename_list:
+            path_validator = InputTypePathMapper[self.input_type.name].value(file)
+
+            source.add(path_validator.instrument)
```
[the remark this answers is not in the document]
maxinelasp: Since SPICEInput overrides this _set_attributes_from_filename method, it shouldn't end up in this situation unless that method is deleted out of SPICEInput.

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,387 @@
...
+            if self.input_type == ProcessingInputType.SCIENCE_FILE:
+                data_type.add(path_validator.data_level)
+            else:
+                data_type.add(self.input_type.value)
+            descriptor.add(path_validator.descriptor)
+            file_path_list.append(path_validator)
```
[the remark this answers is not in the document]
maxinelasp: You're right! This should be `path_validator.filename` instead. I will update.

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,387 @@
...
+    def __init__(self, *args):
+        """Set the processing type to ScienceFile and then calls super().
+
+        Parameters
+        ----------
+        args : str
+            Filenames for initialization.
+        """
+        self.input_type = ProcessingInputType.SCIENCE_FILE
+        super().__init__(*args)
```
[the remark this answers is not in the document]
maxinelasp: Yes, I should have written some tests. I will do so tomorrow. The `super()` call class the `__init__` method in the superclass, which here is defined in ProcessingInput ln 65. Both Science and ancillary files are similar enough to use the same __init__ function, but SPICE will need a different version.

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,387 @@
...
+        day: datetime
+            Input day to retrieve files for
+
+        Returns
+        -------
+        list[str]
+            List of filenames that are required for the given day.
+        """
+        # todo: complete this
+        return NotImplementedError
```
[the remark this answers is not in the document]
maxinelasp: Sure, I can move this up if you think we want it for all the filetypes

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,387 @@
...
+        # Not yet completed
+        raise NotImplementedError
+
+    def _set_attributes_from_filenames(self) -> None:
+        """Set the source, data type, and descriptor attributes based on filename."""
+        # TODO: update SPICEFilePath to retrieve data_type and descriptor from
+        # file name. Do we have an expected filename format?
+
+        # just using examples for now
+        self.source = "sc_attitude"
```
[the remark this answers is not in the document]
maxinelasp: Yes I agree that should be an enum or a class. We should probably move a lot of those classes from sds-data-manager and put it in imap-data-access. I'd personally prefer to leave that for a later PR, but I can do that here if you want.

tech3371:
> @tech3371 are the changes you're requesting just the tests? or do you want me to move those SPICE classes over before merging as well? I feel a lot less confident about SPICE requirements and filenames, which is why that class is a lot more bare than the others...

Looks like SPICE is going to need it's own PR in the future. Lot of my questions and comments were for SPICE. Can you create issue from this PR for future SPICE work? Then we can revisits some of my comments during that time. I think this first pass for science and ancillary is great and it would be very helpful to discuss how we can use that in infrastructure would be helpful offline. For this PR, some tests showing how it would be used together would be enough.

imap_data_access/processing_input_file.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def get_time_range(self):
+        """Retrieve the time range covered by the files.
+
+        Files are assumed to cover exactly 24 hours. The range returned is (earliest,
+        latest) where latest is inclusive.
+
+        Returns
+        -------
+        (earlist, latest) : tuple[datetime]
```
tech3371, line 179: Can you call it (start_time, end_time) like you did in parent class for both of these children class? I just noticed that when I was trying to draw UML diagram to understand this.

maxinelasp:
@tech3371 I added some tests, let me know if there are situations I didn't cover there that you need more info about!

imap_data_access/processing_input.py
```diff
@@ -0,0 +1,385 @@
...
+
+    def get_time_range(self):
+        """Retrieve the time range covered by the files.
+
+        Files are assumed to cover exactly 24 hours. The range returned is (earliest,
+        latest) where latest is inclusive.
+
+        Returns
+        -------
+        (earlist, latest) : tuple[datetime]
```
tech3371, line 179:
```suggestion
        (start_time, end_time) : tuple[datetime]
```

imap_data_access/processing_input.py
```diff
@@ -0,0 +1,385 @@
...
+    def get_time_range(self):
+        """Return the time range covered by the ancillary files.
+
+        The return is a tuple (earliest, latest) where latest is inclusive.
+        For example, a single file with a time range of 20250101-20250105 would return
+        (20250101, 20250105).
+
+        Returns
+        -------
+        (earlist, latest) : tuple[datetime]
```
tech3371, line 225:
```suggestion
        (start_time, end_time) : tuple[datetime]
```

imap_data_access/processing_input.py
tech3371: Can you change these to be consistent with parent function?

imap_data_access/processing_input.py
```diff
@@ -0,0 +1,385 @@
...
+            elif file_creator["type"] == ProcessingInputType.SPICE_FILE.value:
+                self.add(SPICEInput(*file_creator["files"]))
+
+    def get_science_files(self) -> list[ProcessingInput]:
+        """Return just the science files from the collection.
+
+        Returns
+        -------
+        out : list[ProcessingInput]
+            list of ScienceInput files contained in the collection.
```
tech3371, line 378:
```suggestion
            list of ProcessingInput files contained in the collection.
```

tests/test_processing_input.py
```diff
@@ -0,0 +1,126 @@
...
+    assert len(deser.processing_input) == 2
+    assert deser.processing_input[0].descriptor == "l1b-cal"
+    assert deser.processing_input[1].descriptor == "norm-magi"
+    assert deser.processing_input[0].input_type == ProcessingInputType.ANCILLARY_FILE
+    assert deser.processing_input[1].input_type == ProcessingInputType.SCIENCE_FILE
+
+    extra_files = processing_input.ProcessingInputCollection(
+        processing_input.ScienceInput("imap_glows_l1a_hist_20250202_v001.cdf")
+    )
+    assert len(extra_files.processing_input) == 1
```
tech3371, line 103: What is this line checking? Just curious. I would think that line 100 is instantiating new class but seems like it's adding to it. Is that right?

tests/test_processing_input.py
```diff
@@ -0,0 +1,126 @@
...
+def test_get_time_range():
+    ancillary = processing_input.AncillaryInput(
+        "imap_mag_l1b-cal_20250101_v001.cdf",
+        "imap_mag_l1b-cal_20250103-20250104_v002.cdf",
+    )
+
+    start, end = ancillary.get_time_range()
+
+    assert start == datetime.strptime("20250101", "%Y%m%d")
+    assert end == datetime.strptime("20250104", "%Y%m%d")
```
tech3371, line 126: Really nice tests. I understand much more now what this is doing. Thank you so much!

tech3371 (approved):
I had some non-blocking questions and comments. If you can change those, it looks great to me. Thank you for doing this! btw, I hope you are not working today when you are sick.

tests/test_processing_input.py
maxinelasp: Probably a bit of a duplicate test, but just making sure `extra_files` is set up properly before the serialize/de serialize test below here.

imap_data_access/processing_input.py
maxinelasp: Although the type of the return is ProcessingInput, because I can't enforce the type on the return, we are expecting only ScienceInput files (which also count as ProcessingInput files, since ScienceInput is a subclass). So, I'd prefer to leave the comment as-is because it is accurate.

imap_data_access/processing_input.py
maxinelasp: Yep, good call

Issue #123, opened 2025-02-18
Add unit test for validating an AncillaryFilePath


bourque:
We should add a `test_file_validation.test_ancillary_file_path` unit test that makes sure that ancillary files can be validated (similar to `test_file_validation.test_spice_file_path`).

bourque:
Per a bug that I found when working on #124, we should make sure to have a test that doesn't include an `end_date`.

greglucas:
I was trying to add a test for this quick and running into issues. Ancillary files aren't currently allowed to be "csv", but the error message when generating an ancillary file is not helpful. It would be nice to list the allowed extensions if we are going to validate them, or do we want to just yolo with extensions and let them be whatever?

This fails to create a filepath:
```python
    anc_file = AncillaryFilePath.generate_from_inputs(
        instrument="mag",
        description="test",
        start_time="20210101",
        version="v001",
        extension="csv",
    )
```

daralynnrhode:
@greglucas would it be because you have `start_time` instead of `start_date`? 
"csv" is listed within `VALID_ANCILLARY_FILE_EXTENSION` and I couldn't track down a reason why it would fail without knowing what the error message was.

greglucas:
@daralynnrhode, that is a good catch on the inconsistency! It looks like we use `start_time` in the `generate_from_inputs()` function: https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/2feecccbece522f2599662cafee38c7850da5c11/imap_data_access/file_validation.py#L78
But `self.start_date` internally. We should probably correct that too.

https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/main/tests/test_file_validation.py#L213-L219

It looks like AncillaryFilePath subclasses ScienceFilePath, there may be some inheritance creeping in that we don't want or need to override still on AncillaryFilePath? Worth looking at all of this in this issue! Maybe should break some of these out into separate issues so it doesn't become overwhelming to look through this too...

Issue #121, opened 2025-02-17
Update API demo notebook ahead of Feb STM


bourque:
Some things have changed since the last STM and so the demo notebook should be updated to showcase the API's latest capabilities before the Feb STM in case users are interested!

PR #124, opened 2025-02-18
Fixed some small bugs in ancillary file validation step

bourque:
This PR fixes some bugs I discovered while testing the downloading of ancillary files, namely adding in a missing import, and checking for a valid `end_date` only if one is supplied.

### DIFF
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -8,7 +8,11 @@ provides a convenient way to query the IMAP data archive and download data files
 import os
 from pathlib import Path
 
-from imap_data_access.file_validation import ScienceFilePath, SPICEFilePath
+from imap_data_access.file_validation import (
+    AncillaryFilePath,
+    ScienceFilePath,
+    SPICEFilePath,
+)
 from imap_data_access.io import download, query, upload
 
 __all__ = [
@@ -17,6 +21,7 @@ __all__ = [
     "upload",
     "ScienceFilePath",
     "SPICEFilePath",
+    "AncillaryFilePath",
     "VALID_INSTRUMENTS",
     "VALID_DATALEVELS",
     "VALID_FILE_EXTENSION",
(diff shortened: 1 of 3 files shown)


tech3371 (approved):
LGTM

greglucas (commented):
Can you add a test for `validate_filename` with the `self.end_date = None` which I think is what your new addition adds.

bourque:
@greglucas Yes, we should do that, but I would prefer if this would be taken care of in a future PR that solves #123, as I am trying to get this fix in and do some other work before tomorrow's STM.

tech3371:
> @greglucas Yes, we should do that, but I would prefer if this would be taken care of in a future PR that solves #123, as I am trying to get this fix in and do some other work before tomorrow's STM.

Does this fix allow download of ancillary files? Just curious

greglucas:
> @greglucas Yes, we should do that, but I would prefer if this would be taken care of in a future PR that solves https://github.com/IMAP-Science-Operations-Center/imap-data-access/issues/123, as I am trying to get this fix in and do some other work before tomorrow's STM.

I don't know that this has _fixed_ the bug without a test though 🤷 If you found the bug, I think we should add a test personally... This is how we get behind in test coverage and forget to codify things later because adding tests later is so low priority. I won't block over this though.

bourque:
> > @greglucas Yes, we should do that, but I would prefer if this would be taken care of in a future PR that solves #123, as I am trying to get this fix in and do some other work before tomorrow's STM.
> 
> Does this fix allow download of ancillary files? Just curious

@tech3371 Yes. Without this fix, I was getting some errors.

bourque:
> > @greglucas Yes, we should do that, but I would prefer if this would be taken care of in a future PR that solves #123, as I am trying to get this fix in and do some other work before tomorrow's STM.
> 
> I don't know that this has _fixed_ the bug without a test though 🤷 If you found the bug, I think we should add a test personally... This is how we get behind in test coverage and forget to codify things later because adding tests later is so low priority. I won't block over this though.

@greglucas You are welcome to add a test here if you want to get this in now, otherwise I will hold off until we work on #123 (and I will add a note to that issue to make sure to include an `end_date` test). Just let me know what you would like to do, otherwise I will plan to merge this PR in the next few hours so that I can get #121 all set ahead of tomorrow's meeting.

greglucas (approved):

bourque:
Thanks for adding that @greglucas! We will be sure to expand upon this in #123

PR #126, opened 2025-02-18
API demo notebook updates for Feb 2025 STM

### LINKED ISSUE #121: Update API demo notebook ahead of Feb STM

bourque:
This PR updates the `examples/api_demo.ipynb` notebook to showcase the latest version of `imap-data-access` and its latest capabilities; namely adding an example of querying with the `--filename` parameters, and downloading ancillary files.

For reviewers: Unfortunately it is not easy to compare changes in Jupyter notebooks on GitHub. I would suggest taking a look [here](https://github.com/bourque/imap-data-access/blob/api_demo_notebook_updates_feb_2025_STM/examples/api_demo.ipynb) to see what the end product looks like.

Closes #121

### DIFF
--- a/examples/api_demo.ipynb
+++ b/examples/api_demo.ipynb
@@ -2,40 +2,26 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "f28c0487",
-   "metadata": {
-    "collapsed": false,
-    "jupyter": {
-     "outputs_hidden": false
-    },
-    "pycharm": {
-     "name": "#%% md\n"
-    }
-   },
    "source": [
     "# IMAP Data Access API\n",
     "\n",
-    "The IMAP SDC provides a Data Access API to **query**, **download**, and **upload** data. There are three ways to use the API:\n",
+    "The IMAP SDC provides a Data Access API to **query** and **download** data. There are three ways to use the API:\n",
     "\n",
     "1. The `imap-data-access` command-line utility\n",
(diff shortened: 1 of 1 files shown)


greglucas:
I know that last time we had a few questions/concerns over the upload endpoint and who can do that etc... My only minor suggestion is to remove that from the notebook and just show the query/download since that is what end users will be primarily concerned with and we haven't implemented the upload restrictions to our API keys yet. We actually don't want end users using the upload capability in all honesty, we'd prefer to control those if I remember right from our past conversations.

bourque:
Good point Greg. I think before we wanted to be able to show the group "hey look, we can upload data to our processing pipelines", but now that we are pivoting more towards showing users how they can interact with our APIs, it makes sense to nix the upload stuff.

daralynnrhode:
The only critique I have is that the Future Work section is a little outdated. Some of the points have already been implemented like `'latest'` and others were decided to not be introduced, like `N days ago`.

bourque:
Thanks @daralynnrhode, good catch!

bourque:
Could someone please approve this so I can merge it in before tomorrow's SDC update? 🤞🏻

daralynnrhode (approved):
LGTM!

PR #128, opened 2025-03-12
Created factory method for filepath objects

maxinelasp:
# Change Summary
This was inspired by a few discussions and seeing the same code copied around. This method is a convenient way to create `ScienceFilePath` or `AncillaryFilePath` objects.

I also moved some static general methods into a new shared base class (which also acts to make the typehints work.) This shouldn't affect anything else because they're inherited into each class - but it just seemed like a good spot for it. 

May affect work by @lacoak21, @daralynnrhode, and @tech3371, but hopefully to make it simpler!

To use, you call `generate_imap_file_path` on your filename, and then you can use `isinstance` to filter based on file type.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -41,7 +41,7 @@ def generate_imap_file_path(filename: str) -> ImapFilePath:
                 # Matches neither file format
                 error_message = (
                     f"Invalid file type for {filename}. It does not match"
-                    f" Science or Ancillary file formats"
+                    f"Spice, Science or Ancillary file formats"
                 )
                 raise ValueError(error_message) from e
 
@@ -93,22 +93,6 @@ class ImapFilePath:
             Whether input version is valid or not.
         """
         return input_version == "latest" or re.fullmatch(r"v\d{3}", input_version)
-
-    @staticmethod
-    def is_valid_repointing(input_repointing: str) -> bool:
-        """Check input repointing string is in valid format 'repointingXXXXX'.
-
-        Parameters
-        ----------
-        input_repointing : str
-            Repointing to be checked.
-
-        Returns
-        -------
-        bool
-            Whether input repointing is valid or not.
-        """
-        return re.fullmatch(r"repoint\d{5}", str(input_repointing))
 
     @abstractmethod
     def construct_path(self) -> Path:
@@ -360,6 +344,22 @@ class ScienceFilePath(ImapFilePath):
             components["repointing"] = int(components["repointing"])
         return components
 
+    @staticmethod
+    def is_valid_repointing(input_repointing: str) -> bool:
+        """Check input repointing string is in valid format 'repointingXXXXX'.
+
+        Parameters
+        ----------
+        input_repointing : str
+            Repointing to be checked.
+
+        Returns
+        -------
+        bool
+            Whether input repointing is valid or not.
+        """
+        return re.fullmatch(r"repoint\d{5}", str(input_repointing))
+
 
 # Transform the suffix to the directory structure we are using
 # Commented out mappings are not being used on IMAP
@@ -589,7 +589,7 @@ class AncillaryFilePath(ImapFilePath):
         ):
             error_message = (
                 f"Invalid filename, missing attribute. Filename "
-                f"convention is {imap_data_access.FILENAME_CONVENTION} \n"
+                f"convention is {imap_data_access.ANCILLARY_FILENAME_CONVENTION} \n"
             )
         if self.mission != "imap":
             error_message += f"Invalid mission {self.mission}. Please use imap \n"
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -199,8 +199,9 @@ def query(
         raise ValueError("Not a valid version, use format 'vXXX'.")
 
     # check repointing follows 'repoint00000' format
-    if repointing is not None and not file_validation.ImapFilePath.is_valid_repointing(
-        repointing
+    if (
+        repointing is not None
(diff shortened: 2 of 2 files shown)


imap_data_access/file_validation.py
```diff
@@ -4,13 +4,119 @@
...
+                path_obj = imap_data_access.AncillaryFilePath(filename)
+            except AncillaryFilePath.InvalidAncillaryFileError as e:
+                # Matches neither file format
+                error_message = (
+                    f"Invalid file type for {filename}. It does not match"
+                    f" Science or Ancillary file formats"
+                )
+                raise ValueError(error_message) from e
+
+    return path_obj
```
tech3371, line 48: This is great! Another thing I have noticed we tend to do a lot is we check if what data source type is given a filename. Can you think of a way to return that in similar manner as this? I don't know what is the best way for this. For example, to return a `ProcessingInputType:Enum` value or set that value in their respective class. If latter, then we won't need to return additional information and can be referenced from this `path_obj`.  The reason I ask for this is because we do things little different based on input type.

imap_data_access/file_validation.py
```diff
@@ -399,7 +450,7 @@ def construct_path(self) -> Path:
         return spice_dir / subdir / self.filename
 
 
-class AncillaryFilePath(ScienceFilePath):
+class AncillaryFilePath(ImapFilePath):
```
tech3371, line 453: I have noticed that on line `592, it was using `imap_data_access.FILENAME_CONVENTION`. Can we defined a filename convention for ancillary since it's different from science. I got confused few days ago because of the message it returned.

tech3371 (changes requested):
This looks great. I was seeing lot of duplicate code on infrastructure and this is nice feature.

I have two change requests in first and last comment from me on this PR. Besides that, it looks good to me.

imap_data_access/file_validation.py
```diff
@@ -4,13 +4,119 @@
...
+        # Science and Ancillary
+        try:
+            path_obj = imap_data_access.ScienceFilePath(filename)
+        except ScienceFilePath.InvalidScienceFileError:
+            # If Science file fails, then process as an Ancillary file
+            try:
+                path_obj = imap_data_access.AncillaryFilePath(filename)
+            except AncillaryFilePath.InvalidAncillaryFileError as e:
+                # Matches neither file format
+                error_message = (
```
lacoak21, line 42: SPICE should be in the error message too.

lacoak21 (approved):
This is great! generate_imap_file_path can be used in batch_starter.py

imap_data_access/file_validation.py
```diff
@@ -4,13 +4,119 @@
...
+        ----------
+        input_repointing : str
+            Repointing to be checked.
+
+        Returns
+        -------
+        bool
+            Whether input repointing is valid or not.
+        """
+        return re.fullmatch(r"repoint\d{5}", str(input_repointing))
```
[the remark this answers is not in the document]
maxinelasp: Ohh yeah that is true... I'll move it back

imap_data_access/file_validation.py
maxinelasp: Ooh yes good catch!

imap_data_access/io.py
```diff
@@ -201,19 +181,19 @@ def query(
...
     # Check end-date
-    if end_date is not None and not file_validation.ScienceFilePath.is_valid_date(
+    if end_date is not None and not file_validation.ImapFilePath.is_valid_date(
         end_date
     ):
         raise ValueError("Not a valid end date, use format 'YYYYMMDD'.")
 
     # Check version make sure to include 'latest'
-    if version is not None and not file_validation.ScienceFilePath.is_valid_version(
+    if version is not None and not file_validation.ImapFilePath.is_valid_version(
```
tech3371, line 196: Is this change in effort to support both science and ancillary? If so, I think we need to update line 212 to check for ancillary file extension too.

imap_data_access/file_validation.py
tech3371: Luisa told me that we can find input type by doing `isinstance(file_obj, ScienceFilePath)`. Then l will leave it up to you. Thank you for doing this.

imap_data_access/io.py
tech3371: This will require update still.

I noticed something not related to this PR but something we need to work on. Query for ancillary is not working. I will create issue for it.

tech3371 (approved):
If you can make one update as mentioned in the comment, then it looks good to me.

imap_data_access/io.py
tech3371: Created issue #130

daralynnrhode: @tech3371 I am still working on the query support for ancillary files. Got a little blocked when it came to switching between tables (science, spice, or ancillary) with the cli that the changes in this PR should be able to fix. Plus, the upload took higher priority. I'll be getting back to that soon!

tech3371: I understand. Thank you! I will merge PR I have on infrastructure repo to unblock you then. I also noticed that we didn't update ancillary table or we don't write to database of ancillary file ingestion. That's a blocker for ancillary query feature. That PR will clear the road. I will have that merged today or by end of tomorrow.

tech3371: @daralynnrhode I verified that my PR changes works in AWS and just merged that. I wanted to give that update in case you will be waiting on me.

daralynnrhode: @tech3371 Sounds good! Thanks!

PR #129, opened 2025-03-12
Factory function to create processing input from list of files 

lacoak21:
# Change Summary
 
Similarly to what Maxine did in this PR #128 add a factory function to create a processing input from a list of files. This function will be used in batch starter after querying table for upstream dependencies. 

## Updated Files
   - generate_processing_input function
## Testing
Add tests for processing input factory function

### DIFF
--- a/imap_data_access/processing_input.py
+++ b/imap_data_access/processing_input.py
@@ -27,6 +27,45 @@ class InputTypePathMapper(Enum):
     SPICE_FILE = SPICEFilePath
 
 
+def generate_processing_input(filenames: list[str]) -> ProcessingInput:
+    """Generate a ProcessingInput object from a list of filenames.
+
+    This method determines if the filenames are SPICE, Science, or Ancillary files and
+    returns a SPICEInput, ScienceInput, or AncillaryInput object respectively.
+
+    Parameters
+    ----------
+    filenames : list[str]
+        The filenames to generate a ProcessingInput for.
+
+    Returns
+    -------
+    A FilePath object
+    """
+    # Science and Ancillary
+    try:
+        input_obj = ScienceInput(*filenames)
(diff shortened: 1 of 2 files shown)


imap_data_access/processing_input.py
```diff
@@ -27,6 +27,45 @@ class InputTypePathMapper(Enum):
...
+            # except SPICEFilePath.InvalidSPICEFileError:
+
+            # Matches neither input type
+            error_message = (
+                f"Invalid processing inputs for {filenames}. All files much match "
+                f"either SPICE science or ancillary input formats."
+            )
+            raise ValueError(error_message) from e
+
+    return input_obj
```
tech3371, line 66: Is this needed if you get more than one file from database query result? Just speaking out loud

tests/test_processing_input.py
```diff
@@ -124,3 +127,48 @@ def test_get_time_range():
...
+        ),
+    ],
+)
+def test_generate_processing_input(file_list, expected_class):
+    # Test with one file
+    processing_input_one = generate_processing_input(file_list[0:1])
+    assert isinstance(processing_input_one, expected_class)
+    # Test with two files
+    processing_input_two = generate_processing_input(file_list)
+    assert isinstance(processing_input_two, expected_class)
```
tech3371, line 164: Can you write test line that checks if object contains files list you input? Something like this
```
assert processing_input_one.filename_list = [what you expect]
```

tech3371 (approved):
one comment. besides, it looks good to me

lacoak21:
I realized yesterday I do not need this function because when I query each table I should know the data type and therefor which processing input class it should be. I am going to close this.

tech3371:
> I realized yesterday I do not need this function because when I query each table I should know the data type and therefor which processing input class it should be. I am going to close this.

Perfect!

PR #131, opened 2025-03-14
MNT: fixing file path list to have download path

tech3371:
# Change Summary

## Overview
This is to simplify download dependency in general. Greg, Luisa and I discussed some approach and we thought this change may help on `imap_cli` update work. Right now in `cli.py`, we have to go through `filename_list` and then create `construct_path()` and then download each dependency files. Then we loose that information after download. With this change, we create download path list and store in the class and store in its `file_path_list` parameter. Then use that to download the file in `cli.py` and then return `InputCollection()` to all our instrument processing code.  Instrument processing code can use that information to look up their downloaded file from `file_path_list` and also allow them to have other additional informations from `InputCollection()` in processing, such as `input_type` or `descriptor` to decide what to with certain science or ancillary or spice input type.

## Updated Files
   - Change to accommodate above plan

## Testing

### DIFF
--- a/imap_data_access/processing_input.py
+++ b/imap_data_access/processing_input.py
@@ -58,7 +58,7 @@ class ProcessingInput(ABC):
     # List of filenames
     filename_list: list[str] = None
     # List of download paths
-    file_path_list: list[Path] = None
+    file_obj_list: list[AncillaryFilePath, ScienceFilePath, SPICEFilePath] = None
     input_type: ProcessingInputType = None
     # Following three are retrieved from dependency check.
     # But they can also come from the filename.
@@ -110,13 +110,13 @@ class ProcessingInput(ABC):
         This method is called by the constructor and can be overridden by subclasses.
         It works for ScienceFilePaths and AncillaryFilePaths, but not SPICEFilePaths.
 
-        This sets source, datatype, descriptor, and file_path_list attributes.
+        This sets source, datatype, descriptor, and file_obj_list attributes.
         """
         # For science and ancillary files
         source = set()
         data_type = set()
         descriptor = set()
-        file_path_list = []
+        file_obj_list = []
         for file in self.filename_list:
             path_validator = InputTypePathMapper[self.input_type.name].value(file)
 
@@ -126,8 +126,7 @@ class ProcessingInput(ABC):
             else:
                 data_type.add(self.input_type.value)
             descriptor.add(path_validator.descriptor)
-            # Store download path. This gets used in `imap_processing` cli
-            file_path_list.append(path_validator.construct_path())
+            file_obj_list.append(path_validator)
 
         if len(source) != 1 or len(data_type) != 1 or len(descriptor) != 1:
             raise ValueError(
@@ -137,7 +136,7 @@ class ProcessingInput(ABC):
         self.source = source.pop()
         self.data_type = data_type.pop()
         self.descriptor = descriptor.pop()
-        self.file_path_list = file_path_list
+        self.file_obj_list = file_obj_list
 
     def construct_json_output(self):
         """Construct a JSON output.
--- a/imap_data_access/processing_input.py
+++ b/imap_data_access/processing_input.py
@@ -7,9 +7,13 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import Enum
-from pathlib import Path
 
-from imap_data_access import AncillaryFilePath, ImapFilePath, ScienceFilePath, SPICEFilePath
+from imap_data_access import (
+    AncillaryFilePath,
+    ImapFilePath,
+    ScienceFilePath,
+    SPICEFilePath,
+)
 
 
 class ProcessingInputType(Enum):
--- a/imap_data_access/processing_input.py
+++ b/imap_data_access/processing_input.py
@@ -42,8 +42,10 @@ class ProcessingInput(ABC):
 
     Attributes
     ----------
-    filepath_list : list[str]
-        A list of filepaths.
+    filename_list : list[str]
+        A list of filename(s).
+    file_obj_list : list[AncillaryFilePath, ScienceFilePath, SPICEFilePath]
+        A list of file objects, one for each filename.
     input_type : ProcessingInputType
         The type of input file.
     source : str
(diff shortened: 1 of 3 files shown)


tech3371:
pre-commit.ci autofix

imap_data_access/processing_input.py
```diff
@@ -54,7 +55,10 @@ class ProcessingInput(ABC):
         A descriptor for the file, for example, "burst" or "cal".
     """
 
+    # List of filenames
```
greglucas, line 58: Do we need this list of filenames? Aren't these taken care of by Path.name of your list items below?

If you do need this, could it be a property that maps over the file paths instead?

I thought we discussed storing the ScienceFilePath classes we have in these lists, but it looks like it is a plain Path object without the specific classes, or maybe I'm misunderstanding the return types of some of these functions or where we discussed doing that somewhere else instead.

lacoak21: I also am confused about this. I thought file_path_list would be a list of the ScienceFilePaths.

imap_data_access/processing_input.py
```diff
@@ -122,7 +126,8 @@ def _set_attributes_from_filenames(self):
             else:
                 data_type.add(self.input_type.value)
             descriptor.add(path_validator.descriptor)
-            file_path_list.append(str(path_validator.filename))
+            # Store download path. This gets used in `imap_processing` cli
+            file_path_list.append(path_validator.construct_path())
```
lacoak21, line 130: Here I thought it was going to be:
`file_path_list.append(path_validator)`

imap_data_access/processing_input.py
tech3371: Oh interesting, then I must have misunderstood during that conversation. Sorry.

imap_data_access/processing_input.py
tech3371: I can do that. Looks like that's what we discussed and I misunderstood. Then with this, we will need to change the name. Any suggestions?

And in `cli.py` on `imap_processing`, I can loop through this list and call `construct_path()`. That's minor change on that end and works for me.

imap_data_access/processing_input.py
maxinelasp: With the [new changes](https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/128), you can now set this to be a list of ImapFilePath objects to accomplish what Luisa and Greg are saying here. I would also say that we should remove `filepath_list` and instead add an abstract method to `ImapFilePath` that says `get_filename` so we are only storing the info in one place. I'll add some other comments describing this with proposed changes.

imap_data_access/processing_input.py
```diff
@@ -54,7 +55,10 @@ class ProcessingInput(ABC):
         A descriptor for the file, for example, "burst" or "cal".
     """
 
+    # List of filenames
     filename_list: list[str] = None
+    # List of download paths
+    file_path_list: list[Path] = None
```
maxinelasp, line 61:
```suggestion
    imap_file_paths: list[ImapFilePath] = None
```

imap_data_access/processing_input.py
maxinelasp, line 130: as Luisa said: 
```suggestion
            imap_file_paths.append(path_validator)
```

Here, path_validator will be `ScienceFilePath` or `AncillaryFilePath` which are both `ImapFilePath` types, so that should work.

maxinelasp: I can't comment on it, but then also in ln 148 you would need to use `self.imap_file_paths.get_filename()`. These changes mean we're storing `ScienceFilePath` objects with all those convenient functions.

maxinelasp (changes requested):
Some notes on filenames that I think accomplish everything in Luisa's comments.

imap_data_access/processing_input.py
```diff
@@ -41,8 +46,10 @@ class ProcessingInput(ABC):
 
     Attributes
     ----------
-    filepath_list : list[str]
-        A list of filepaths.
+    filename_list : list[str]
```
maxinelasp, line 49: I would still prefer to have only imap_file_paths rather than having imap_file_paths and filename_list, but that's not a super-strong preference. Otherwise looks good!

lacoak21 (approved):
Changes look good to me Tenzin!

imap_data_access/processing_input.py
tech3371: I can't take `filename_list` out because it's used by `construct_json_output()` and we need `imap_file_paths` for everything else. I think they both are doing something in their own way. I don't feel strongly about it too. I will see if I can reduce one parameter.

tech3371: @maxinelasp I looked at removing `filename_list` but it's used pretty heavily. In my last commit, I did refactor code to remove that. Can you see if it's something you want or if I should undo that last commit? Based on your response, I can remove this doc string as well and then merge it.

imap_data_access/processing_input.py
```diff
@@ -108,20 +106,25 @@ def get_time_range(self):
...
+        filename_list : list[str]
+            List of input filenames to set attributes.
         """
         # For science and ancillary files
         source = set()
         data_type = set()
         descriptor = set()
         file_obj_list = []
-        for file in self.filename_list:
+        for file in filename_list:
```
maxinelasp, line 127: Rather than taking in the filename lists, you can loop through self.imap_file_paths and remove the line that creates path_validator.

tests/test_processing_input.py
```diff
@@ -18,18 +18,19 @@ def test_create_science_files():
         "imap_mag_l1a_burst-magi_20240310_v000.cdf",
     )
 
-    assert one_file.filename_list == ["imap_mag_l1a_norm-magi_20240312_v000.cdf"]
+    assert one_file.construct_json_output()["files"] == ["imap_mag_l1a_norm-magi_20240312_v000.cdf"]
```
maxinelasp, line 21: Instead, should be `assert [file.filename in one_file.imap_file_paths] == ["imap_mag_l1a_norm-magi_20240312_v000.cdf"]` so we're specifically testing the construction of the class.

maxinelasp (commented):
Some comments on the filename_list stuff. If you prefer to leave in filename_list and revert your last commit, I'm also fine with that!

tech3371:
pre-commit.ci autofix

maxinelasp (approved):

PR #132, opened 2025-03-17
MNT: Refactor ImapFilePath to be a Path object itself

greglucas:
# Change Summary

## Overview

Per a discussion we had last week about trying to make a `ScienceFilePath` an actual `Path` and whether that was possible. I think this will simplify our lives and make it a `Path` just like it is in our minds by the name already.

This lets us use the Path related methods directly on our Paths and eliminates the need for `construct_path()`. It means the resolved path is what we want as well and we don't need to store self.data_dir and filename anymore.

I've added in the methods/properties with deprecation warnings for backwards compatibility.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -3,10 +3,11 @@
 
 from __future__ import annotations
 
+import os
 import re
-from abc import abstractmethod
+import warnings
 from datetime import datetime
-from pathlib import Path
+from pathlib import Path, PosixPath, WindowsPath
 
 import imap_data_access
 
@@ -48,7 +49,9 @@ def generate_imap_file_path(filename: str) -> ImapFilePath:
     return path_obj
 
 
-class ImapFilePath:
+# TODO: In Python3.12+ we can subclass Path directly
+#       for now we need to handle posix and windows separately
+class ImapFilePath(WindowsPath if os.name == "nt" else PosixPath):
     """Base class for FilePaths.
 
     Includes shared static methods and provides correct typing for ScienceFilePath,
@@ -94,10 +97,37 @@ class ImapFilePath:
         """
         return input_version == "latest" or re.fullmatch(r"v\d{3}", input_version)
 
-    @abstractmethod
     def construct_path(self) -> Path:
         """Construct valid path from class variables and data_dir."""
-        raise NotImplementedError
+        warnings.warn(
+            "This method is deprecated and will be removed in a future"
+            "release. Use the current instance as a Path object instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.resolve()
+
+    @property
+    def data_dir(self) -> Path:
+        """Get the data directory from the config."""
+        warnings.warn(
+            "This method is deprecated and will be removed in a future"
+            "release. Use imap_data_access.config['DATA_DIR'] instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return imap_data_access.config["DATA_DIR"]
+
+    @property
+    def filename(self) -> str:
+        """Return the filename of the path."""
+        warnings.warn(
+            "This method is deprecated and will be removed in a future"
+            "release. Use Path.name instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.name
 
 
 class ScienceFilePath(ImapFilePath):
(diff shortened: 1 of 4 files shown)


imap_data_access/file_validation.py
```diff
@@ -48,7 +49,9 @@ def generate_imap_file_path(filename: str) -> ImapFilePath:
     return path_obj
 
 
-class ImapFilePath:
+# TODO: In Python3.12+ we can subclass Path directly
+#       for now we need to handle posix and windows separately
+class ImapFilePath(WindowsPath if os.name == "nt" else PosixPath):
```
maxinelasp, line 54: This piece seems fragile and that's a little concerning to me...

maxinelasp (commented):
Pros: simplifies things, I do think it makes sense, adds functionality 
Cons: Adds a lot of complexity on the types, which is scary to me (because Python is not very good with types), changes API of this class, since this is in imap-data-access I am concerned that the wheels/build step and the WindowsPath vs PosixPath will cause problems (i.e. if imap_processing is built on Windows and run on Unix, will all the Python wheels and OS steps resolve correctly?) 

So, overall I'm a little neutral on this PR. I think the drawbacks are probably worth it for the benefits, but there are a few pieces that are a little worrying to me, and I don't see an easy fix for any of them. But I also don't see any obvious places where those concerns would pop up, so maybe it's fine. 

I think I would feel better if `ImapFilePath` inherited directly from the "pure" Path class and then overwrote the dunder method(s) of the class and selected from PosixPath or WindowsPath at that point, but maybe that's way more janky and wouldn't really work. 

Something like this: 
```
from pathlib import Path as _Path_, PosixPath, WindowsPath

class Path(_Path_):
 def __new__(cls, *args):
  if cls is Path:
    cls = WindowsPath if os.name == 'nt' else PosixPath
setattr(cls, "extramethod", _extramethod)
return cls._from_parts(args)
```

 
from [option 3 here](https://codereview.stackexchange.com/questions/162426/subclassing-pathlib-path) is vaguely what I'm thinking about...

greglucas:
> Cons: Adds a lot of complexity on the types, which is scary to me (because Python is not very good with types), changes API of this class, since this is in imap-data-access I am concerned that the wheels/build step and the WindowsPath vs PosixPath will cause problems (i.e. if imap_processing is built on Windows and run on Unix, will all the Python wheels and OS steps resolve correctly?)

The class type is changed at runtime based on the platform that is detected. There is no compilation going on for us because we are pure Python.

> I think I would feel better if ImapFilePath inherited directly from the "pure" Path class and then overwrote the dunder method(s) of the class and selected from PosixPath or WindowsPath at that point, but maybe that's way more janky and wouldn't really work.

I like the BasePath-like inheritance with the if-else based on platform type instead of injecting this logic into the 3 separate `__new__` versions.

NOTE: I got nerd-sniped on this one... Apparently Python significantly changed the subclassing behavior in Python 3.12, so before that we need to use `__new__` to set everything up for our subclasses, but 3.12 and later we need to use `__init__` to properly set things up. This means we need to implement both and also put a version gate on the `__init__`. I factored out the logic so we could share things across both methods across all 3 implementations.

This is definitely making things more complicated in the class instantiation, so we should weigh whether this is valuable for later on. My thought was that it would make some of https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/131 easier to handle each `ScienceFilePath` because you have an actual `Path` everywhere. Later we can do `imap_data_access.download(my_science_file_path)` and things will "just work" without having to call `.name` / `.filename` etc...

imap_data_access/file_validation.py
greglucas: I don't think this is fragile, this is just an unfortunate situation with older versions of Python not directly allowing a subclass of Path, so here we have to decide beforehand which one to choose and that happens at runtime here depending on which system we are on.

maxinelasp (changes requested):
Having thought about it more, I don't think I like this change. All the different requirements for different versions and operating systems is concerning to me. Unfortunately Path (when below 3.13) just is difficult to subclass and I don't think the benefits of saving a few function calls are worth the added complexity.

maxinelasp:
I also don't really see where it would help, most of the time we are passing around these ScienceFilePaths and only use it as a `Path` at the last second.

maxinelasp:
These classes were created because we were using `Path` objects all over the place and it required a lot of validation checks because we're expecting a super specific kind of object (which lines up to the filename definitions exactly.) So they're intended to be very specific. I think moving to a more generic `Path` inheritance just complicates things again and invites people to use methods that we don't really want them to use anyway. It also only saves one occasional function call and makes these classes a lot more complicated, so I don't think I understand the benefit.

greglucas:
> These classes were created because we were using Path objects all over the place and it required a lot of validation checks because we're expecting a super specific kind of object (which lines up to the filename definitions exactly.) So they're intended to be very specific.

I think that first comment illustrates why I thought this was useful. I'm saying lets get back to using Path objects, but these Path objects now have validation upon instantiation.

I'm also just iffy on this and there are some weird things when you subclass. So I'll close it for now and think a bit more about it.

PR #133, opened 2025-03-17
Adding Ancillary File Path Test(s)

### LINKED ISSUE #123: Add unit test for validating an AncillaryFilePath

daralynnrhode:
# Adding Tests to cover Ancillary File Validation

## Overview

Per the bugs found in #123 , adding tests to validation ancillary file paths to avoid them. 
Closes #123 

Turns out it was simple miss spelling error because csv is not the same as cvs.

## New Files

- 

## Deleted Files

- 

## Updated Files

- test_file_validation.py

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -666,7 +666,7 @@ class AncillaryFilePath(ImapFilePath):
             r"(?P<start_date>\d{8})"
             r"(-(?P<end_date>\d{8}))?"  # Optional end_date field
             r"_(?P<version>v\d{3})"
-            r"\.(?P<extension>cdf|cvs|json)$"
+            r"\.(?P<extension>cdf|csv|json)$"
         )
         if isinstance(filename, Path):
             filename = filename.name
--- a/tests/test_file_validation.py
+++ b/tests/test_file_validation.py
@@ -208,6 +208,58 @@ def test_spice_file_path():
     )
 
 
+def test_ancillary_file_path_all_parameters():
+    """Tests the ``construct_path`` method with no end_date provided."""
+    anc_file = AncillaryFilePath.generate_from_inputs(
+        instrument="mag",
+        descriptor="test",
+        start_time="20210101",
+        end_time="20210102",
+        version="v001",
+        extension="cdf",
+    )
+
+    assert anc_file.validate_filename() == ""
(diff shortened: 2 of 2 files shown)


imap_data_access/file_validation.py
```diff
@@ -666,7 +666,7 @@ def extract_filename_components(filename: str | Path) -> dict:
             r"(?P<start_date>\d{8})"
             r"(-(?P<end_date>\d{8}))?"  # Optional end_date field
             r"_(?P<version>v\d{3})"
-            r"\.(?P<extension>cdf|cvs|json)$"
+            r"\.(?P<extension>cdf|csv|json)$"
```
greglucas, line 669: Good catch!

tests/test_file_validation.py
```diff
@@ -208,18 +208,72 @@ def test_spice_file_path():
...
         instrument="mag",
         descriptor="test",
         start_time="20210101",
+        end_time="20210102",
         version="v001",
         extension="cdf",
     )
-
-    assert anc_file.validate_filename() == ""
+    assert ancillary_file_all_params.validate_filename() == ""
```
greglucas, line 233: This is already taken care of within instantiation, where we only make the ancillary file if it is valid. So if we get here without an error we know this filename is valid. I think it could be dropped, but it also doesn't hurt too much, just feels a bit odd to be comparing to an empty string to me.

greglucas (approved):
Looks great, thanks for adding these!

bourque:
> In #123 @greglucas mentioned adding a way to list the allowed extensions for file types, is that something that is still needed? Maybe some output for the validation that prints what portion it fails on and what the values should be, something like `Invalid Ancillary File, must follow this format <file_convention_ancillary> \n Error with Instrument section: Must be <instruments>`

@daralynnrhode That would be helpful! If you want to implement something like that feel free. Probably best to open a different PR for that since this one is almost wrapped up.

bourque (approved):
Agree with Greg's comment about not needing the `assert validate_filename` bit. Otherwise this looks good to me!  Nice work. It is good that we now have tests for different scenarios.

PR #138, opened 2025-03-21
ENH: Add webpoda download capability to the project

greglucas:
# Change Summary

## Overview

We need to download L0 data from the POC. This brings the script I shared in Slack into our imap-data-access repository so we can use it across the project and keep all the io-related tasks in one repository. This helps to download the data to `IMAP_DATA_DIR` and generates the L0 filepath based on the ScienceFile requirements.

This is a separate URL on-premises. It requires authentication so we add a new `IMAP_WEBPODA_TOKEN` environment variable to get that externally.

Add a new cli subcommand to query webpoda specifically. Test with the following command.

```
imap-data-access webpoda --instrument swapi --start-date 20241202
```

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -16,12 +16,14 @@ Use
 """
 
 import argparse
+import datetime
 import logging
 import os
 from pathlib import Path
 
 import imap_data_access
 from imap_data_access.file_validation import ScienceFilePath
+from imap_data_access.webpoda import download_daily_data
 
 
 def _download_parser(args: argparse.Namespace):
@@ -174,6 +176,29 @@ def _upload_parser(args: argparse.Namespace):
     print("Successfully uploaded the file to the IMAP SDC")
 
 
+def _webpoda_parser(args: argparse.Namespace):
+    """Download raw packet data from IMAP.
+
+    Parameters
+    ----------
+    args : argparse.Namespace
+        An object containing the parsed arguments and their values
+    """
+    if args.end_date:
+        end_time = args.end_date
+    else:
+        end_time = args.start_date
+    # Now push that out to 23:59:59
+    end_time = datetime.datetime.combine(end_time, datetime.time.max)
+
+    download_daily_data(
+        instrument=args.instrument,
+        start_time=args.start_date,
+        end_time=end_time,
+    )
+    print("Successfully downloaded the data from webpoda.")
+
+
 # PLR0915: too many statements
 def main():  # noqa: PLR0915
     """Parse the command line arguments.
@@ -184,6 +209,11 @@ def main():  # noqa: PLR0915
         "API key to authenticate with the IMAP SDC. "
         "This can also be set using the IMAP_API_KEY environment variable. "
         "It is only necessary for uploading files."
+    )
+    webpoda_token_help = (
+        "Used to authenticate with the IMAP Project. "  # noqa: S105
+        "This can also be set using the IMAP_WEBPODA_TOKEN environment variable. "
+        "It is only necessary for downloading binary packet data."
     )
     data_dir_help = (
         "Directory to use for reading and writing IMAP data. "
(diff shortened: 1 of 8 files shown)


imap_data_access/webpoda.py
```diff
@@ -0,0 +1,348 @@
...
+        f"between earth received time {start_time} and {end_time}"
+    )
+    logger.info(f"Unique spacecraft dates with packets: {unique_dates}")
+
+    # Iterate over the packet dates to make a query for each individual spacecraft day
+    # packet_date 00:00:00 -> packet_date 23:59:59
+    # TODO: Do we always want this buffer?
+    # The buffer is used to make sure we get packets grouped together appropriately
+    # even if they cross the day boundaries
+    buffer = datetime.timedelta(minutes=30)
```
maxinelasp, line 312: We could add a buffer kwarg to allow individual instruments to turn it on or off

maxinelasp (approved):
Nice, this looks good.

imap_data_access/webpoda.py
greglucas: Great suggestion! I actually added this as a default argument and then removed it because I thought it was trying to add too many things at once and some people wanted different buffers on either end and I wasn't sure how this would interact with that if we do need the capability.

For now, I prefer to keep this simple and fixed internally, but will definitely open up a ticket for this idea. Thanks!

greglucas: Yep! #140 for the buffer, and #141 for the repointing

imap_data_access/cli.py
```diff
@@ -174,6 +176,29 @@ def _upload_parser(args: argparse.Namespace):
...
+    ----------
+    args : argparse.Namespace
+        An object containing the parsed arguments and their values
+    """
+    if args.end_date:
+        end_time = args.end_date
+    else:
+        end_time = args.start_date
+    # Now push that out to 23:59:59
+    end_time = datetime.datetime.combine(end_time, datetime.time.max)
```
vmartinez-cu, line 192: Good to know that `datetime.time.max` gives the latest time available in a day. That's useful

imap_data_access/webpoda.py
```diff
@@ -0,0 +1,348 @@
...
+def download_daily_data(
+    instrument: str,
+    start_time: datetime.datetime,
+    end_time: datetime.datetime,
+    version: str = "v001",
+    upload_to_server=False,
+):
+    """Download data for the apid and start/end time range from webpoda.
+
+    PODA stands for packet on demand access. This function requests the IMAP specific
```
vmartinez-cu, line 268: This is a helpful docstring. I didn't know what PODA stood for until just now

vmartinez-cu (approved):
LGTM

imap_data_access/webpoda.py
```diff
@@ -0,0 +1,348 @@
...
+    instrument : str
+        The instrument to download data for.
+    start_time : datetime.datetime
+        The start time of the query in Earth Received Time (ERT).
+    end_time : datetime.datetime
+        The end time of the query in Earth Received Time (ERT).
+    version : str, optional
+        The version to use on the downloaded data file, by default "v001"
+    upload_to_server : bool, optional
+        If True, upload the data to the server, by default False
```
tech3371, line 285: ?
```suggestion
        If True, upload the data to the data bucket, by default False
```

imap_data_access/webpoda.py
tech3371: Does today SDC tag-up discussion result in removing this buffer?

tech3371 (approved):
Thank you so much for this. The code was very easy to follow thanks to great doc strings and variable names!!!

imap_data_access/webpoda.py
greglucas: Updated!

imap_data_access/webpoda.py
greglucas: Yes, I just removed this in the latest push.

@maxinelasp, the SPDF does not want duplicate data as indicated in their e-mail response, so we've decided to remove the buffer here (per the SDC tag-up this morning). I think this means that we will need to account for the overlapping data periods on MAG in some other way. I'll follow up on this via e-mail and we can add any buffering/other info necessary later as well.

PR #142, opened 2025-03-21
ENH: Add ialirt and spacecraft as valid instruments

greglucas:
`ialirt` and `spacecraft` aren't actual instruments, but they are additional data sources for packet definitions and processing. So we need to be able to specify these on the command lines and through validation functions.


tech3371 (approved):
LGTM

PR #147, opened 2025-03-26
Spice files

bryan-harter:
# Change Summary

## Overview
Adding in file name parsing for SPICE files

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
   - Added more strict regex to the SPICE files 
   - Added an "extract" method that will return the spice type/version information
   - Added a new mapping dictionary to convert the types in the file names into something more readable
- tests/test_file_validation.py
   - Added unit tests to ensure the "extract_parts" method worked
   - Updated a few unit tests to use real SPICE file names
- tests/test_io.py
   - Updated a unit test to use real SPICE file names
   - Updated a unit test to use real SPICE file names

## Testing
The tests already cover a large portion of the SPICEFilePath class, but I added tests to make sure we could extract file type/version information from all of the different SPICE files we could expect.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -519,7 +519,7 @@ class SPICEFilePath(ImapFilePath):
             SPICE data filename or file path.
         """
         self.filename = Path(filename)
-        self.spice_metadata = self.extract_filename_components()
+        self.spice_metadata = SPICEFilePath.extract_filename_components(self.filename, ["type", "version"])
         if self.spice_metadata is None:
             raise self.InvalidSPICEFileError(
                 f"Invalid SPICE file. Expected file to have one of the following "
@@ -544,9 +544,9 @@ class SPICEFilePath(ImapFilePath):
         # Use the file suffix to determine the directory structure
         # IMAP_DATA_DIR/spice/<subdir>/filename
         return spice_dir / subdir / self.filename
-
+    @staticmethod
     def _matches_on_group(
-        self, regex_list: Iterable[re.Pattern], string_to_parse: str
+        regex_list: Iterable[re.Pattern], string_to_parse: str
     ) -> re.Match:
         """Determine the first regular expression that matches the provided string.
 
@@ -566,9 +566,10 @@ class SPICEFilePath(ImapFilePath):
             if m is not None:
                 return m
         return None
-
+    
+    @staticmethod
     def _extract_parts(
-        self,
+        filename: Path | str,
         parts: list[str],
         transforms: dict | None = None,
         handle_missing_parts: bool = False,
@@ -577,6 +578,8 @@ class SPICEFilePath(ImapFilePath):
 
         Arguments:
         ---------
+            filename: Path | str
+                A filename to extract parts from
             parts: list[str]
                 A list of parts to be extracted from the filename
             transforms: dict[str: func]
@@ -592,8 +595,11 @@ class SPICEFilePath(ImapFilePath):
             A dictionary of the parts that were requested or
             None if there were no matches for the provided regex_list
         """
-        m = self._matches_on_group(
-            SPICEFilePath.valid_spice_regexes, self.filename.name.lower()
+        if isinstance(filename, Path):
+            filename = filename.name
+
+        m = SPICEFilePath._matches_on_group(
+            SPICEFilePath.valid_spice_regexes, filename.lower()
         )
         if not m:
             return None
@@ -615,7 +621,9 @@ class SPICEFilePath(ImapFilePath):
 
         return ret_val
 
-    def extract_filename_components(self) -> dict | None:
+    @staticmethod
+    def extract_filename_components(filename: Path | str,
+                                    parts: list[str] = ['type', 'version']) -> dict | None:
         """Extract all components from filename.
 
         Will return a dictionary with the following keys:
@@ -623,13 +631,21 @@ class SPICEFilePath(ImapFilePath):
 
         If a match is not found, None will be returned
 
+        Parameters
+        ----------
+            filename: Path | str
+                The filename to parse
+            parts: list[str]
+                The name of the sections to extract
+
         Returns
         -------
         components : dict
             Dictionary containing components.
         """
-        spice_metadata = self._extract_parts(
-            ["type", "version"],
+        spice_metadata = SPICEFilePath._extract_parts(
+            filename,
+            parts,
             transforms={
                 "type": lambda x: _SPICE_TYPE_MAPPING.get(x, None),
                 "version": lambda x: int(x),
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import re
 from abc import abstractmethod
+from collections.abc import Iterable
 from datetime import datetime
 from pathlib import Path
 
@@ -430,15 +431,14 @@ class SPICEFilePath(ImapFilePath):
     # Spin Files (type: spin.csv)
     # Repoint Files (type: repoint.csv)
     attitude_file_pattern = (
-        r"(?P<{mission}>imap)_"
-        r"(?P<{start_year}>[\d]{{4}})_"
-        r"(?P<{start_doy}>[\d]{{3}})_"
-        r"(?P<{end_year}>[\d]{{4}})_"
-        r"(?P<{end_doy}>[\d]{{3}})_"
-        r"(?P<{version}>[a-zA-Z0-9\-_]+)\."
-        r"(?P<{type}>ah.bc|ap.bc|spin.csv|repoint.csv)"
+        r"(?P<mission>imap)_"
+        r"(?P<start_year>[\d]{4})_"
+        r"(?P<start_doy>[\d]{3})_"
+        r"(?P<end_year>[\d]{4})_"
+        r"(?P<end_doy>[\d]{3})_"
+        r"(?P<version>[a-zA-Z0-9\-_]+)\."
+        r"(?P<type>ah.bc|ap.bc|spin.csv|repoint.csv)"
     )
-
     # Covers:
     # Reconstructed (type: recon)
     # Nominal (type: nom)
@@ -447,12 +447,12 @@ class SPICEFilePath(ImapFilePath):
     # Long Term Predict (type: long)
     # Launch Predict (type: launch)
     spacecraft_ephemeris_file_pattern = (
-        r"(?P<{mission}>imap)_"
-        r"(?P<{type}>[a-zA-Z0-9\-]+)_"
-        r"(?P<{start_date}>[\d]{{8}})_"
-        r"(?P<{end_date}>[\d]{{8}})"
-        r"(?:|_v(?P<{version}>[\d]*))\."
-        r"(?P<{extension}>bsp)"
+        r"(?P<mission>imap)_"
+        r"(?P<type>[a-zA-Z0-9\-]+)_"
+        r"(?P<start_date>[\d]{8})_"
+        r"(?P<end_date>[\d]{8})"
+        r"(?:|_v(?P<version>[\d]*))\."
+        r"(?P<extension>bsp)"
     )
 
     # Covers:
@@ -461,48 +461,42 @@ class SPICEFilePath(ImapFilePath):
     # Leapsecond kernel (type: "naif")
     # Spacecraft clock kernel (type: "imapsclk_")
     spice_prod_ver_pattern = (
-        r"(?P<{type}>[a-zA-Z\-_]+)"
-        r"(?P<{version}>[\d]+)\."
-        r"(?P<{extension}>tls|tpc|bsp|tsc)"
+        r"(?P<type>[a-zA-Z\-_]+)"
+        r"(?P<version>[\d]+)\."
+        r"(?P<extension>tls|tpc|bsp|tsc)"
     )
 
     # Covers:
     # Frame: (type: 'tf')
-    spice_frame_pattern = (
-        r"(?P<{mission}>imap)_" r"(?P<{version}>[\d]+)\." r"(?P<{type}>tf)"
-    )
+    spice_frame_pattern = r"(?P<mission>imap)_" r"(?P<version>[\d]+)\." r"(?P<type>tf)"
 
     # Covers:
     # Thruster files (type: sff)
     sff_filename_pattern = (
-        r"(?P<{mission}>imap)_"
-        r"(?P<{start_year}>[\d]{{4}})_"
-        r"(?P<{start_doy}>[\d]{{3}})_"
-        r"(?P<{mode}>[a-zA-Z0-9\-_]+)_"
-        r"(?P<{version}>[\d]{{2}})\."
-        r"(?P<{type}>sff)"
+        r"(?P<mission>imap)_"
+        r"(?P<start_year>[\d]{4})_"
+        r"(?P<start_doy>[\d]{3})_"
+        r"(?P<mode>[a-zA-Z0-9\-_]+)_"
+        r"(?P<version>[\d]{2})\."
+        r"(?P<type>sff)"
     )
 
     # Covers:
     # Metakernels (type: 'tm')
     mk_filename_pattern = (
-        r"(?P<{mission}>imap)_"
-        r"(?P<{start_year}>[\d]{{4}})_"
-        r"v(?P<{version}>[\d]{{3}})\."
-        r"(?P<{type}>tm)"
+        r"(?P<mission>imap)_"
+        r"(?P<start_year>[\d]{4})_"
+        r"v(?P<version>[\d]{3})\."
+        r"(?P<type>tm)"
     )
 
-    valid_spice_regexes = re.compile(
-        "|".join(
-            [
-                attitude_file_pattern,
-                spacecraft_ephemeris_file_pattern,
-                spice_prod_ver_pattern,
-                spice_frame_pattern,
-                sff_filename_pattern,
-                mk_filename_pattern,
-            ]
-        )
+    valid_spice_regexes = (
+        re.compile(attitude_file_pattern),
+        re.compile(spacecraft_ephemeris_file_pattern),
+        re.compile(spice_prod_ver_pattern),
+        re.compile(spice_frame_pattern),
+        re.compile(sff_filename_pattern),
+        re.compile(mk_filename_pattern),
     )
 
     class InvalidSPICEFileError(Exception):
@@ -525,9 +519,7 @@ class SPICEFilePath(ImapFilePath):
             SPICE data filename or file path.
         """
         self.filename = Path(filename)
-        self.spice_metadata = SPICEFilePath.extract_filename_components(
-            self.filename, ["type", "version"]
-        )
+        self.spice_metadata = SPICEFilePath.extract_filename_components(self.filename)
         if self.spice_metadata is None:
             raise self.InvalidSPICEFileError(
                 f"Invalid SPICE file. Expected file to have one of the following "
@@ -554,6 +546,29 @@ class SPICEFilePath(ImapFilePath):
         return spice_dir / subdir / self.filename
 
     @staticmethod
+    def _matches_on_group(
+        regex_list: Iterable[re.Pattern], string_to_parse: str
+    ) -> re.Match:
+        """Determine the first regular expression that matches the provided string.
+
+        Parameters
+        ----------
+        regex_list: list
+            A list of compiled regular expressions to be applied in order
+        string_to_parse: str
+                The string to check against the provided regular expressions
+
+        Returns
+        -------
+            The first match that satisfies a regular expression found in regex_list
+        """
+        for regex in regex_list:
+            m = regex.match(string_to_parse)
+            if m is not None:
+                return m
+        return None
+
+    @staticmethod
     def _extract_parts(
         filename: Path | str,
         transforms: dict | None = None,
@@ -576,7 +591,9 @@ class SPICEFilePath(ImapFilePath):
         if isinstance(filename, Path):
             filename = filename.name
 
-        m = SPICEFilePath.valid_spice_regexes.match(filename.lower())
+        m = SPICEFilePath._matches_on_group(
+            SPICEFilePath.valid_spice_regexes, filename.lower()
+        )
         if m is None:
             return None
         ret_val = m.groupdict()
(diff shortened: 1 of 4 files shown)


imap_data_access/file_validation.py
```diff
@@ -400,6 +425,80 @@ def is_valid_repointing(input_repointing: str) -> bool:
...
+        r"(?P<{}>imap)_"
+        r"(?P<{}>[\d]{{4}})_"
+        r"(?P<{}>[\d]{{3}})_"
+        r"(?P<{}>[\d]{{4}})_"
+        r"(?P<{}>[\d]{{3}})_"
+        r"(?P<{}>[a-zA-Z0-9\-_]+)\."
+        r"(?P<{}>ah.bc|ap.bc|spin.csv|repoint.csv)"
+    ).format(
+        "mission", "year_start", "doy_start", "year_end", "doy_end", "version", "type"
+    )
```
greglucas, line 443: On all of these regexes can we avoid the `.format()` call? It is hard to trace with the parenthesis what gets filled in where.

I think we've been doing `start_date` / `start_time` rather than `date_start`, so swapping the order of the names.
```suggestion
        r"(?P<{mission}>imap)_"
        r"(?P<{start_year}>[\d]{{4}})_"
        r"(?P<{start_doy}>[\d]{{3}})_"
        r"(?P<{end_year}>[\d]{{4}})_"
        r"(?P<{end_doy}>[\d]{{3}})_"
        r"(?P<{version}>[a-zA-Z0-9\-_]+)\."
        r"(?P<{type}>ah.bc|ap.bc|spin.csv|repoint.csv)"
    )
```

imap_data_access/file_validation.py
```diff
@@ -400,6 +425,80 @@ def is_valid_repointing(input_repointing: str) -> bool:
...
+    ).format("mission", "year", "version", "type")
+
+    valid_spice_regexes = (
+        re.compile(attitude_file_pattern),
+        re.compile(spacecraft_ephemeris_file_pattern),
+        re.compile(spice_prod_ver_pattern),
+        re.compile(spice_frame_pattern),
+        re.compile(sff_filename_pattern),
+        re.compile(mk_filename_pattern),
+    )
```
greglucas, line 500: Do you want to combine them into a single match? I think you can do this as a large regex with the pipe operator.
```suggestion
    valid_spice_regexes = re.compile("|".join([attitude_file_pattern, spacecraft_ephemeris_file_pattern, spice_prod_ver_pattern, spice_frame_pattern, sff_filename_pattern, mk_filename_pattern)
```

imap_data_access/file_validation.py
```diff
@@ -444,11 +542,122 @@ def construct_path(self) -> Path:
...
         """
         spice_dir = imap_data_access.config["DATA_DIR"] / "spice"
-        subdir = _SPICE_DIR_MAPPING[self.file_extension]
+        subdir = _SPICE_DIR_MAPPING[self.spice_metadata["type"]]
         # Use the file suffix to determine the directory structure
         # IMAP_DATA_DIR/spice/<subdir>/filename
         return spice_dir / subdir / self.filename
 
+    @staticmethod
+    def _matches_on_group(
```
greglucas, line 551: I think if you combine all the regexes with "|", then you can just call `regex.match()` on that directly and don't need this function.

imap_data_access/file_validation.py
```diff
@@ -444,11 +542,122 @@ def construct_path(self) -> Path:
...
+            set will contain None for the missing part.
+            If False, an IndexError is raised
+
+        Returns
+        -------
+            A dictionary of the parts that were requested or
+            None if there were no matches for the provided regex_list
+        """
+        if isinstance(filename, Path):
+            filename = filename.name
```
greglucas, line 602:
```suggestion
       filename = Path(filename).name
```

imap_data_access/file_validation.py
```diff
@@ -444,11 +542,122 @@ def construct_path(self) -> Path:
...
+
+        m = SPICEFilePath._matches_on_group(
+            SPICEFilePath.valid_spice_regexes, filename.lower()
+        )
+        if not m:
+            return None
```
greglucas, line 608:
```suggestion
        if m is None:
            return None
```

imap_data_access/file_validation.py
```diff
@@ -444,11 +542,122 @@ def construct_path(self) -> Path:
...
+
+        Parameters
+        ----------
+        filename: Path | str
+            A filename to extract parts from
+        parts: list[str]
+            A list of parts to be extracted from the filename
+        transforms: dict[str: func]
+            A dictionary of group->function (that takes 1 string) to
+            transform the string into some other type
```
greglucas, line 590: This could be inlined I think rather than passed here? It looks like you pass a lambda later on, but could this just be fixed internally instead?

imap_data_access/file_validation.py
```diff
@@ -444,11 +542,122 @@ def construct_path(self) -> Path:
...
+        handle_missing_parts: bool
+            If True, missing parts won't raise an exception and the result
+            set will contain None for the missing part.
+            If False, an IndexError is raised
```
greglucas, line 594: Is this unused? Can we remove it if so.

imap_data_access/file_validation.py
```diff
@@ -444,11 +542,122 @@ def construct_path(self) -> Path:
...
+
+        return ret_val
+
+    @staticmethod
+    def extract_filename_components(
+        filename: Path | str, parts: list[str]
+    ) -> dict | None:
+        """Extract all components from filename.
+
+        Will return a dictionary with the following keys:
```
greglucas, line 633: All that we are getting is type and version? What about all of the start_year and other parts that were extracted via the regex. (I think we'd prefer to not pass in parts explicitly and let it fall out automatically based on what was matched?)

I'm a bit lost in the regex handling here, so some examples might be helpful of an input and expected return in one or two specific cases.

In the other functions in this file we regex match and extract out the components directly. Does that not work with the multiple regex pattern?

imap_data_access/file_validation.py
```diff
@@ -400,6 +425,80 @@ def is_valid_repointing(input_repointing: str) -> bool:
...
+    # Predicted Attitude (type: ap.bc)
+    # Spin Files (type: spin.csv)
+    # Repoint Files (type: repoint.csv)
+    attitude_file_pattern = (
+        r"(?P<{}>imap)_"
+        r"(?P<{}>[\d]{{4}})_"
+        r"(?P<{}>[\d]{{3}})_"
+        r"(?P<{}>[\d]{{4}})_"
+        r"(?P<{}>[\d]{{3}})_"
+        r"(?P<{}>[a-zA-Z0-9\-_]+)\."
```
tech3371, line 439: Does this allow version to be any format? Just curious

imap_data_access/file_validation.py
```diff
@@ -400,6 +425,80 @@ def is_valid_repointing(input_repointing: str) -> bool:
...
+        r"(?P<{}>[a-zA-Z0-9\-_]+)_"
+        r"(?P<{}>[\d]{{2}})\."
+        r"(?P<{}>sff)"
+    ).format("mission", "year", "doy_start", "mode", "version", "type")
+
+    # Covers:
+    # Metakernels (type: 'tm')
+    mk_filename_pattern = (
+        r"(?P<{}>imap)_" r"(?P<{}>[\d]{{4}})_" r"v(?P<{}>[\d]{{3}})\." r"(?P<{}>tm)"
+    ).format("mission", "year", "version", "type")
```
tech3371, line 491: Is this the one you/we are producing?

imap_data_access/file_validation.py
bryan-harter: Yeah I definitely think that looks better, but for some reason our other projects use the format thing so I was just copying those. Maybe so you can quickly see what fields are in the regex?

imap_data_access/file_validation.py
bryan-harter: Yup!

imap_data_access/file_validation.py
```diff
@@ -420,16 +519,15 @@ def __init__(self, filename: str | Path):
...
         """
         self.filename = Path(filename)
-        if self.filename.suffix == ".csv":
-            all_suffixes = self.filename.suffixes  # Returns ['.spin', '.csv']
-            self.file_extension = "".join(all_suffixes)  # Returns '.spin.csv'
-        else:
-            self.file_extension = self.filename.suffix
-
-        if self.file_extension not in _SPICE_DIR_MAPPING:
+        self.spice_metadata = SPICEFilePath.extract_filename_components(
```
tech3371, line 522: should this be `self.`? For a second I thought this was different class. If so, can you change in other places as well.
```suggestion
        self.spice_metadata = self.extract_filename_components(
```

imap_data_access/file_validation.py
greglucas: > Alternatively, I could just have the extract_filename_components return anything and everything it can based on the file name, and get rid of the specific "parts" parameter

Yes, I think this would be ideal. Then add in your additional derived type/version as necessary to those found parts.

tests/test_file_validation.py
```diff
@@ -197,17 +197,87 @@ def test_spice_file_path():
...
+            "type",
+        ],
+    )
+    assert file_parts["mission"] == "imap"
+    assert file_parts["year_start"] == "2025"
+    assert file_parts["doy_start"] == "032"
+    assert file_parts["year_end"] == "2025"
+    assert file_parts["doy_end"] == "034"
+    assert file_parts["version"] == 3
+    assert file_parts["type"] == "attitude_history"
```
tech3371, line 278: you can test extract part on some other unique ones you had in `valid_spice_regexes`? This test was helpful to understand the code.

tech3371 (approved):
SPICE filename conventions is so different and unique. Your solutions was also very unique and easy to follow. Nice!

imap_data_access/file_validation.py
bryan-harter: Actually it appears that the problem with a single match is that it complains that I have too many fields named "mission", "version", "type", etc, once  they all get combined into one big regex

imap_data_access/file_validation.py
bryan-harter: Yeah I could do that rather than pass it in. I was trying to be flexible but I don't necessarily know under what scenario you'd ever want to transform a name into something different

tests/test_file_validation.py
bryan-harter: Sure, I can add more tests for each type of file we expect

imap_data_access/file_validation.py
bryan-harter: Yeah it does, I think that is a mistake actually, so I just changed it to only digits

imap_data_access/file_validation.py
```diff
@@ -400,6 +425,80 @@ def is_valid_repointing(input_repointing: str) -> bool:
...
+    spice_frame_pattern = r"(?P<mission>imap)_" r"(?P<version>[\d]+)\." r"(?P<type>tf)"
+
+    # Covers:
+    # Thruster files (type: sff)
+    sff_filename_pattern = (
+        r"(?P<mission>imap)_"
+        r"(?P<start_year>[\d]{4})_"
+        r"(?P<start_doy>[\d]{3})_"
+        r"(?P<mode>[a-zA-Z0-9\-_]+)_"
+        r"(?P<version>[\d]{2})\."
```
maxinelasp, line 480: Should version be 3 digits?

imap_data_access/file_validation.py
```diff
@@ -444,11 +540,101 @@ def construct_path(self) -> Path:
...
+        for part in ret_val:
+            if transforms is not None and part in transforms:
+                ret_val[part] = transforms[part](m.group(part))
+            else:
+                ret_val[part] = m.group(part)
+
+        return ret_val
+
+    @staticmethod
+    def extract_filename_components(filename: Path | str) -> dict | None:
```
maxinelasp, line 608: I don't think there is any validation done in these methods - could you add some to confirm the correct version structure, that the date is valid, etc?

maxinelasp (commented):
Looks good! As a note, there is also a processing input object that can now be updated, so I'm happy to see this change come through. I don't have much info on if the SPICE structure looks correct but I think the code is good!

imap_data_access/file_validation.py
```diff
@@ -400,6 +425,80 @@ def is_valid_repointing(input_repointing: str) -> bool:
 class SPICEFilePath(ImapFilePath):
     """Class for building and validating filepaths for SPICE files."""
 
+    # Covers:
+    # Historical Attitude (type: ah.bc)
+    # Predicted Attitude (type: ap.bc)
+    # Spin Files (type: spin.csv)
+    # Repoint Files (type: repoint.csv)
```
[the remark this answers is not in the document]
bryan-harter: Ok sounds good, I just added that regex

imap_data_access/file_validation.py
bryan-harter: We haven't received a sample file yet, but looking at https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/221734667/SDC+Expected+SPICE+File+List it appears there is only 2 digits?

imap_data_access/file_validation.py
```diff
@@ -400,6 +425,80 @@ def is_valid_repointing(input_repointing: str) -> bool:
...
+    # Spacecraft clock kernel (type: "imapsclk_")
+    spice_prod_ver_pattern = (
+        r"(?P<type>[a-zA-Z\-_]+)"
+        r"(?P<version>[\d]+)\."
+        r"(?P<extension>tls|tpc|bsp|tsc)"
+    )
+
+    # Covers:
+    # Frame: (type: 'tf')
+    spice_frame_pattern = r"(?P<mission>imap)_" r"(?P<version>[\d]+)\." r"(?P<type>tf)"
```
greglucas, line 479:
```suggestion
    spice_frame_pattern = r"(?P<mission>imap)_"(?P<version>[\d]+)\."(?P<type>tf)"
```

imap_data_access/file_validation.py
greglucas: Ahh, that makes sense. Thanks for checking!

imap_data_access/file_validation.py
```diff
@@ -420,16 +519,13 @@ def __init__(self, filename: str | Path):
...
         self.filename = Path(filename)
-        if self.filename.suffix == ".csv":
-            all_suffixes = self.filename.suffixes  # Returns ['.spin', '.csv']
-            self.file_extension = "".join(all_suffixes)  # Returns '.spin.csv'
-        else:
-            self.file_extension = self.filename.suffix
-
-        if self.file_extension not in _SPICE_DIR_MAPPING:
+        self.spice_metadata = SPICEFilePath.extract_filename_components(self.filename)
+        if self.spice_metadata is None:
```
greglucas, line 532: Maybe have `extract_filename_components` raise this error instead since it is invalid there too.

imap_data_access/file_validation.py
greglucas: I prefer to limit options if they are unused, so definitely suggest removing parts and transforms here and do it all.

I'd also suggest removing the _match_on_groups and inline that here since this is the only place it is used I think?

imap_data_access/file_validation.py
maxinelasp: Sounds good! Well, sounds acceptable, I don't love that it's randomly different, but thanks for checking!

imap_data_access/file_validation.py
```diff
@@ -400,6 +425,89 @@ def is_valid_repointing(input_repointing: str) -> bool:
...
+        r"(?P<end_doy>[\d]{3})_"
+        r"(?P<version>[\d]+)\."
+        r"(?P<type>ah.bc|ap.bc|spin.csv)"
+    )
+    # Covers:
+    # Repoint Files (type: repoint.csv)
+    repoint_file_pattern = (
+        r"(?P<mission>imap)_"
+        r"(?P<start_year>[\d]{4})_"
+        r"(?P<start_doy>[\d]{3})_"
```
tech3371, line 446:
```suggestion
        r"(?P<end_year>[\d]{4})_"
        r"(?P<end_doy>[\d]{3})_"
```

tests/test_file_validation.py
tech3371: thank you!

bryan-harter:
Alrighty, I think I responded to everyone's comments and made the necessary changes!

imap_data_access/file_validation.py
bryan-harter: Added some checks in this function

imap_data_access/file_validation.py
```diff
@@ -99,6 +99,33 @@ def construct_path(self) -> Path:
         """Construct valid path from class variables and data_dir."""
         raise NotImplementedError
 
+    @staticmethod
+    def is_valid_doy(input_doy: str, input_year: str) -> bool:
```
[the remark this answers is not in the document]
bryan-harter: Ah you're right, I wrote the function expecting that I'd use it, but didn't!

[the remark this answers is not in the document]
bryan-harter: But actually thats a good idea, I think we can combine them into one field, I don't think that a DOY is ever present without the year right before it in any of the valid regexes

bryan-harter:
I just changed "extract_filename_components" so that it always returns a dictionary of the form:

"version" : int
"type": str
"extension": str
"start_date": datetime
"end_date": datetime

If start_date and/or end_date can't be determined from the file name, they will automatically be assigned a date that spans from the earliest possible date to the end of the data in "de440.bsp" (some time in the year 2650). This is what we would need to insert into our database anyway that way those columns aren't blank for those files. 

I could also make it not return 'start_date' or 'end_date' if they can't be determined, and just let the indexer handle it. I just thought it might be nice to have a dictionary with a consistent format.

imap_data_access/file_validation.py
```diff
@@ -444,11 +539,142 @@ def construct_path(self) -> Path:
...
         """
         spice_dir = imap_data_access.config["DATA_DIR"] / "spice"
-        subdir = _SPICE_DIR_MAPPING[self.file_extension]
+        subdir = _SPICE_DIR_MAPPING[self.spice_metadata["type"]]
         # Use the file suffix to determine the directory structure
         # IMAP_DATA_DIR/spice/<subdir>/filename
         return spice_dir / subdir / self.filename
 
+    @staticmethod
+    def _handle_doy(part, components):
```
greglucas, line 548: This was confusing to me at first because we are modifying objects in-place that were passed in (including deleting them), catching errors and returning strings. There are a lot of moving pieces within this function that I think could get inconsistent very easily if errors happen partway through.

I think this all could be much simpler if we do the parsing of the date on the value in-line below. Let the datetime ValueErrors propagate up if they happen.

imap_data_access/file_validation.py
```diff
@@ -444,11 +539,142 @@ def construct_path(self) -> Path:
...
+        for regex in SPICEFilePath.valid_spice_regexes:
+            m = regex.match(filename.name.lower())
+            if m is not None:
+                spice_metadata = SPICEFilePath._spice_parts_handler(m.groupdict())
+                spice_metadata["extension"] = filename.suffix[1:]
+                if "start_date" not in spice_metadata:
+                    spice_metadata["start_date"] = SPICEFilePath.EARLIEST_VALID_TIME
+                if "end_date" not in spice_metadata:
+                    spice_metadata["end_date"] = SPICEFilePath.LATEST_VALID_TIME
+                break
```
greglucas, line 666: Outside this loop you can handle the bare raise_case without needing the if then.
```suggestion
                return spice_metadata
```

imap_data_access/file_validation.py
```diff
@@ -444,11 +539,142 @@ def construct_path(self) -> Path:
...
+        parts_list = list(components.keys())
+        transforms = {
+            "start_year_doy": SPICEFilePath._handle_doy,
+            "end_year_doy": SPICEFilePath._handle_doy,
+            "start_date": SPICEFilePath._handle_date,
+            "end_date": SPICEFilePath._handle_date,
+            "type": SPICEFilePath._handle_type,
+            "start_year": SPICEFilePath._handle_start_year,
+            "version": SPICEFilePath._handle_version,
+        }
```
greglucas, line 621: My suggestion is to not be too fancy and just manually handle these here. I personally don't think we need to report multiple error strings and the like. That can be follow up if people really like super-validation, but sometimes just failing multiple times is fine IMO, especially if it simplifies the overall code structure.

```suggestion
        if "start_date" in components:
            components["start_date"] = datetime.strptime(components["start_date"], "%Y%m%d")
        if "start_year_doy" in components:
            components["start_date"] = datetime.strptime(components.pop("start_year_doy"), "%Y_%j")
```

greglucas (changes requested):
I personally think this is getting too fancy and confusing to follow and we should try to stick with minimal in imap-data-access and leave more complex validation/needs to the other packages.

imap_data_access/file_validation.py
bryan-harter: I initially did that in-line, but the pre-commit checks complained I had too many if statements lol. I'll see if I can reduce the number

greglucas: Don't feel like we have to obey our precommit overlords. If it makes sense to you to do that, then feel free to add a noqa block and say why you chose that.

tests/test_file_validation.py
```diff
@@ -192,22 +193,150 @@ def test_spice_file_path():
...
+    )
+    assert file_path.spice_metadata["extension"] == "bsp"
+    assert len(file_path.spice_metadata) == 5
+
+
+def test_spice_extract_repoint_parts():
+    file_path = SPICEFilePath("imap_2025_230_01.repoint.csv")
+    assert file_path.spice_metadata["version"] == "01"
+    assert file_path.spice_metadata["type"] == "repoint"
+    assert file_path.spice_metadata["extension"] == "csv"
```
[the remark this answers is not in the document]
tech3371: Good point. What's there is good. I have been thinking about this after email discussion with Ian. I think with this new repoint file change Ian and I discussed, we don't need `repoint.csv` anymore. The latest and greatest repointing file will do what we originally thought about SDC maintained repoint file which was that `repoint.csv` you were thinking. Between that new repoint file Ian and I discussed, and the SPICE ingest table Bryan will work on later, that will get us enough information we need.

PR #148, opened 2025-03-26
MNT: Allow for all files to be returned from input collection

greglucas:
get_file_paths() would raise on both options being None, but we can simplify this to return all paths in that case, which can then be used in the download function to simplify that.


tech3371 (approved):
LGTM

PR #150, opened 2025-03-28
MNT: WEBPODA added error handling, corrected APIDS

maxinelasp:
# Change Summary
Nice API! Unfortunately it didn't run for MAG because WEBPODA was missing some APIDs. I assume in that case we want to continue running but log an error, so I added some error handling to make that happen. I also removed unneeded APIDs for MAG and GLOWS. 

## Overview

- Added error handling around returns 
- Updated APIDs

### DIFF
--- a/imap_data_access/webpoda.py
+++ b/imap_data_access/webpoda.py
@@ -21,10 +21,10 @@ import urllib.request
 import urllib.response
 
 import imap_data_access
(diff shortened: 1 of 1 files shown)


greglucas:
Closing this as the relevant changes happened in #152. Feel free to re-open/update this PR if there is more work to be done though.

PR #152, opened 2025-03-31
FIX: update GLOWS and MAG apids

### LINKED ISSUE #151: BUG - Can't download MAG data
### Description of the issue

There is currently an APID that doesn't exist in MAG, which returns a 404 response from the webpoda server.

```
INFO:imap_data_access.webpoda:Downloading data for instrument [mag]
HTTP Error: 404 - Not Found
Server Message: Resource not found: /apids/SID2/apid_1036.txt
```

```
imap-data-access -v webpoda --instrument mag --start-date 20241202
```

We should not be trying to download invalid apids, or catch this case and warn a user but continue on to download the other potential apids.

All processing stops because an exception occurred.

### Code Snippet:

```python
Code
```

_No response_

greglucas:
Some of the apids listed in GLOWS and MAG were incorrect and causing failures when downloading. This only updates the apids and doesn't add the additional try/except from #150.

closes #151 

Tested all instruments with the following script and all apids are at least valid and don't fail the processing anymore:
```

for instrument in INSTRUMENT_APIDS:
    print(instrument)
    # Download the data for the instrument
    download_daily_data(
        instrument,
        datetime.datetime(2008, 12, 2, 0, 0, 0),
        datetime.datetime(2008, 12, 2, 23, 59, 59),
        version="v001",
        upload_to_server=False,
    )
```


imap_data_access/webpoda.py
```diff
@@ -55,12 +55,12 @@
...
         1174,
     ],
-    "glows": [1445, 1448, 1449, 1480, 1481],
+    "glows": [1480, 1481],
     "hi": [754, 769, 770, 818, 833, 834, 818, 833, 834],
     "hit": [1251, 1252, 1253],
     "idex": [1377, 1420, 1424],
     "lo": [673, 676, 677, 705, 706, 707, 708],
-    "mag": [1000, 1036, 1052, 1063, 1064, 1068, 1082],
+    "mag": [1052, 1068],
```
tech3371, line 63: During CLI update, we noticed that MAG didn't produced burst L1A files. Is that because of missing APID or is that expect right now @maxinelasp?

tech3371 (approved):
The code looks good but I don't know enough about what APID MAG or GLOWS needs.

imap_data_access/webpoda.py
greglucas: I assume so. Were you running into this issue?
https://github.com/IMAP-Science-Operations-Center/imap-data-access/issues/151

maxinelasp: @tech3371 Not all the provided test runs include burst data. It's ok for the pipeline to only have one, right now it should work up through L1C

PR #156, opened 2025-04-04
ENH: Add a method to download repointing data

### LINKED ISSUE #141: Add ability to download from webpoda based on repointing time
### Summary

The current webpoda download is a daily file, which works well for in-situ data products. However, ENA teams want data grouped by repointing times.

We will need to add another function to query based on repointing and group data based on those times.

### Requirements

  - [ ] Calculate which repointings had data updated after the Earth received times. We are currently doing this based on what "day" the S/C time falls into. We'd need to group by "repointing" instead. This may require external knowledge to be passed in.
  - [ ] Create the packet files with the repointing number in the filename as well.

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview

It turns out the ENA teams do want data downloaded based on repointing, so this adds the capability described in #141.

This queries by ERT, but then groups based on repointing rather than by day. The repointing knowledge is external and requires the repointing file to be passed in.

closes #141

### DIFF
--- a/imap_data_access/webpoda.py
+++ b/imap_data_access/webpoda.py
@@ -403,15 +403,24 @@ def download_repointing_data(  # noqa: PLR0913
     # Iterate over the packet dates to make a query for each individual "pointing"
     # A "pointing" is defined as the time between the end of one repointing maneuver
     # to the end of the next repointing maneuver.
+    # NOTE: We iterate over the repointings rather than the packet times because it is
+    #       assumed to be the shorter list (1/day vs 1000s of packets/day per apid)
     for i in range(len(repointings) - 1):
         pointing_start = datetime.datetime.strptime(
             repointings[i]["repoint_end_time_utc"], "%Y-%m-%dT%H:%M:%S.%f"
         )
         if repointings[i + 1]["repoint_end_time_utc"].lower() == "nan":
+            # Missing repointing end time, so it isn't a complete "pointing" yet.
+            continue
+        if pointing_start > packet_times[-1]:
+            # This pointing is after the last packet time, so skip it
             continue
         pointing_end = datetime.datetime.strptime(
             repointings[i + 1]["repoint_end_time_utc"], "%Y-%m-%dT%H:%M:%S.%f"
         )
+        if pointing_end < packet_times[0]:
+            # This pointing is before the first packet time, so skip it
+            continue
         if not any(pointing_start <= p_time <= pointing_end for p_time in packet_times):
             # This pointing didn't contain any packets within it
             continue
--- a/imap_data_access/webpoda.py
+++ b/imap_data_access/webpoda.py
@@ -14,11 +14,13 @@ Location of list of APIDs and associated instruments:
 https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/155648242/Packet+Decommutation+Resource+Page+-+IMAP
 """
 
+import csv
 import datetime
 import logging
 import urllib.parse
 import urllib.request
 import urllib.response
+from pathlib import Path
 
 import imap_data_access
 from imap_data_access.io import _get_url_response
@@ -294,9 +296,6 @@ def download_daily_data(
     ]
 
     # Get the unique dates from the packet times
-    # TODO: Account for repointing times and filter out packets between repointings?
-    #       Filtering out at this point seems dangerous, we should probably try to
-    #       filter in the instrument processing stage instead.
     unique_dates = set([dt.date() for dt in packet_times])
     logger.info(
         f"Found [{len(packet_times)}] packets for instrument [{instrument}] "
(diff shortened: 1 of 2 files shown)


imap_data_access/webpoda.py
```diff
@@ -340,3 +339,115 @@ def download_daily_data(
...
+            repoint_start_time_utc	str
+            repoint_end_time_utc	str
+            repoint_id	UINT
+    version : str, optional
+        The version to use on the downloaded data file, by default "v001"
+    upload_to_server : bool, optional
+        If True, upload the data to the SDC data bucket, by default False
+    """
+    # Store a list of rows in the repointing file
+    with open(repointing_file) as f:
```
maxinelasp, line 384: Not relevant now, but for the long term, were you picturing using some kind of repointing API for this? Or pointing to the database?

imap_data_access/webpoda.py
```diff
@@ -340,3 +339,115 @@ def download_daily_data(
...
+    )
+    logger.info(
+        f"Found [{len(packet_times)}] packets for instrument [{instrument}] "
+        f"between earth received time {start_time} and {end_time}"
+    )
+
+    # Iterate over the packet dates to make a query for each individual "pointing"
+    # A "pointing" is defined as the time between the end of one repointing maneuver
+    # to the end of the next repointing maneuver.
+    for i in range(len(repointings) - 1):
```
maxinelasp, line 406: Does it make sense to loop through the packets rather than the repointings here? Or limit the start time to the earliest time in the packet? This list will get bigger as we go. Also more a note for future refactors, I think, as for right now it's completely fine and probably will be fine forever 😅

maxinelasp (approved):
LGTM!

imap_data_access/webpoda.py
greglucas: This file is the "database" now I believe, so I don't think there is going to be an API other than one that says "get me the latest repointing file". I haven't fully kept up though, so this might be something to consider in the future.

This is a great point that we need a way to get this file still. I am addressing that in the sds-data-manager, but that doesn't help here or with a potential CLI. I'll write a ticket up for this and I think it might be doable with querying the API to find the latest repointing file and then downloading that automatically for the user without them having to do multiple steps. (I don't know if we have the query portion available yet for this either...)

imap_data_access/webpoda.py
greglucas: Great suggestion, thanks!

I added a note, we should be having much fewer pointings compared to packets I believe which is why I decided to iterate over the repointings. Basically we need to group the packets into repointings similar to the group by days in t he above function to find the unique days/pointings. I couldn't think of a great way immediately to do that in a nicer way, so suggestions are welcome if there is something we can do to identify which pointings have data better than this.

I did also add a few checks for the first/last packet vs pointing start/end which should exit the loop faster now if the pointing is out of the range of sorted packets.

maxinelasp: Barely related, but it would also be good to add some verification we aren't losing packets doing this style...

greglucas: > Barely related, but it would also be good to add some verification we aren't losing packets doing this style...

Thank you! I had been meaning to address this. We were actually including a packet right on a repointing boundary into both files because our subsequent queries are <= and >=. I've now subtracted one second from the end and indicated that a repointing is [start, end) now so we don't include that packet in both files. Test added with a packet file right on the boundary.

maxinelasp: Doesn't that mean if a packet is .5 seconds before the end of the boundary we will miss it? We should also floor the repointing to one second to account for that case.

greglucas: No, that is already handled by the queries being by `%Y-%m-%dT%H:%M:%S`, so our highest resolution in the query with the equals signs is one-second.

subagonsouth: This seems like a reasonable implementation to me. Worst case, we are iterating over most of the packet times int this if: `if not any(pointing_start <= p_time <= pointing_end for p_time in packet_times):`. If the packets are organized by aipd rather than sorted by time in the `packet_times` list, that should significantly cut down on the likelyhood that all the packets get checked.

subagonsouth (commented):
LGTM. I will get Paul's opinion on the definition of pointing that I unilaterally decided on here.

imap_data_access/webpoda.py
```diff
@@ -340,3 +339,128 @@ def download_daily_data(
...
+    packet_times = sorted(
+        [p for apid in apids for p in get_packet_times_ert(apid, start_time, end_time)]
+    )
+    logger.info(
+        f"Found [{len(packet_times)}] packets for instrument [{instrument}] "
+        f"between earth received time {start_time} and {end_time}"
+    )
+
+    # Iterate over the packet dates to make a query for each individual "pointing"
+    # A "pointing" is defined as the time between the end of one repointing maneuver
```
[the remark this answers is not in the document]
greglucas: Sounds great. I think the main difference will be whether we use start_time / end_time of the maneuver lookups here. I don't have a major preference whatsoever and your comment the other convinced me this was slightly better to do end-time of a maneuver as the start of a pointing (I think I remember that is what we decided on long ago as well for the definition of when a pointing starts).

Sounds good Maxine, I think that lines up here. This is giving you the full pointing+maneuver data, and then each instrument will filter out the maneuver on their own so that we aren't dropping/missing any packets in delivery and then everyone can do what they wish with those (including ignoring them).

greglucas:
> I will get Paul's opinion on the definition of pointing that I unilaterally decided on here.

Sounds good. I am happy to update this later / change definitions. Getting this in now so we can (attempt to) use it during MSIM.

PR #158, opened 2025-04-04
FIX: Handle empty packet returns and add more logging

greglucas:
While testing this during a time with no packets on an instrument. I noticed that we failed with a ValueError for indexing into `packet_times[-1]` when `packets_times` was empty. This catches that error and gracefully returns now instead.

I also noticed that one of the logs didn't have an f-string added for apid so added that above as well.


subagonsouth (approved):
LGTM

PR #161, opened 2025-04-07
MNT/FIX: Update SPICE sclk expected filename mapping

### LINKED ISSUE #160: BUG - The imap SCLK kernel has changed naming conventions
### Description of the issue

The new kernel is named `imap_sclk_####.tsc` (with an underscore, whereas previously it was concatenated `imapsclk`).
https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/c9e056000cb3e7bf199d57a225249706a4ca2dfd/imap_data_access/file_validation.py#L465-L469

We need to update this to use an underscore, and also check whether there are other downstream changes this impacts like in `imap_processing`

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

greglucas:
# Change Summary

## Overview

This updates the expected filename convention for the sclk kernel to use `imap_sclk_####.tsc` I updated the type-mapping portion rather than adding a new/different regex. Updating the regex directly caused naif and other filenames to fail. Not sure if this is the most robust approach or if others want another regex for this specifically.

closes #160

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -379,7 +379,7 @@ _SPICE_TYPE_MAPPING = {
     "de": "planetary_ephemeris",
     "pck": "planetary_constants",
     "naif": "leapseconds",
-    "imapsclk_": "spacecraft_clock",
+    "imap_sclk_": "spacecraft_clock",
     "tf": "frames",
     "tm": "metakernel",
     "sff": "thruster",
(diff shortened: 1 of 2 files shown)


greglucas:
Moving to draft until we get confirmation from the MOC on which version it is supposed to be. There is some conflicting repsonses/versions currently.

bryan-harter (approved):

tech3371 (approved):
Based on Mac's response, we are going with this change. LGTM

PR #163, opened 2025-04-07
 Fix IDEX APIDs

lacoak21:
# Change Summary

## Overview
Remove extraneous APIDS. The IDEX team has confirmed this is the only APID needed by SDC. 

## Updated Files
- imap_data_access/webpoda.py
   - Edit APIDs


greglucas (approved):
Looks good! We discussed following up with IDEX to see whether they want those other apids decommutated or not, but they can be brought back in again later if so and would require some rework on imap_processing side to accomodate.

greglucas:
Getting this in so we don't miss the next release. We can add them back in later if need be.

PR #164, opened 2025-04-07
FIX: Webpoda downloads shouldn't be impacted by duplicate uploads

greglucas:
# Change Summary

## Overview

When duplicate uploads occur to the IMAP SDC, we would throw an IMAPDataAccessError which would propagate up and stop all other download/uploads for other days of this instrument, or other instruments when used in a loop. We don't necessarily need to raise this error, but can rather just log it as an error and move on attempting all of the other days too.

Note that we have to do this within this inner loop if we get two days worth of data and one of them was duplicated but the other wasn't. There is no way to catch that in the outer instrument loop. Say 2025/04/03 was already present, and our webpoda client tries to create both 2025/04/03 and 2025/04/04 this next time.

This likely needs some additional future thought on how to address the duplicate versions (bump version automatically, something smarter, ...), but that could be slotted into this same code area later after discussions. This is an attempt to get things continuing on for now in our current automatic downloads.

### DIFF
--- a/imap_data_access/webpoda.py
+++ b/imap_data_access/webpoda.py
@@ -23,7 +23,7 @@ import urllib.response
 from pathlib import Path
 
 import imap_data_access
-from imap_data_access.io import _get_url_response
+from imap_data_access.io import IMAPDataAccessError, _get_url_response
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


greglucas:
Ping for a review on this if we are going to be updating sds-data-manager so we can start getting more data flowing during MSIMs.

subagonsouth:
> This likely needs some additional future thought on how to address the duplicate versions (bump version automatically, something smarter, ...), but that could be slotted into this same code area later after discussions. This is an attempt to get things continuing on for now in our current automatic downloads.

My thought is that we need to check a checksum against the existing file to see if there is any difference in the data. If the new data is different, it would indicate that data was added because the WebPODA query would presumably have been over the same time range. In that case, we want to increment the version and upload. If for some weird reason, the checksum is identical, there is no new data.

Thinking this through actually has me worried that the implementation in this ticket will mean for MSIM3, we will never generate a new file for a day or pointing which means that partial days will never be filled.

PR #165, opened 2025-04-07
Adding MOC metakernel to the list of expected kernels

tech3371:
# Change Summary

## Overview
Adding MOC attitude and ephemeris metakernel to the list.

## Updated Files
   - changes to add it to the metakernels

## Testing
- tests/test_file_validation.py

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -486,8 +486,8 @@ class SPICEFilePath(ImapFilePath):
 
     # Covers:
     # SDC generated metakernels (type: 'tm')
-    mk_filename_pattern = (
-        r"(imap)_"
+    sdc_mk_filename_pattern = (
+        r"(imap)_sdc_metakernel_"
         r"(?P<start_year>[\d]{4})_"
         r"v(?P<version>[\d]{3})\."
         r"(?P<type>tm)"
@@ -515,7 +515,7 @@ class SPICEFilePath(ImapFilePath):
         re.compile(spice_prod_ver_pattern),
         re.compile(spice_frame_pattern),
         re.compile(sff_filename_pattern),
-        re.compile(mk_filename_pattern),
+        re.compile(sdc_mk_filename_pattern),
         re.compile(attitude_mk_filename_pattern),
         re.compile(ephemeris_mk_filename_pattern),
     )
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -368,6 +368,7 @@ class ScienceFilePath(ImapFilePath):
 _SPICE_TYPE_MAPPING = {
     "ah.bc": "attitude_history",
     "ap.bc": "attitude_predict",
+    "mk": "metakernel",
     "spin.csv": "spin",
     "repoint.csv": "repoint",
     "recon": "ephemeris_reconstructed",
@@ -484,12 +485,27 @@ class SPICEFilePath(ImapFilePath):
     )
 
     # Covers:
-    # Metakernels (type: 'tm')
+    # SDC generated metakernels (type: 'tm')
     mk_filename_pattern = (
         r"(imap)_"
         r"(?P<start_year>[\d]{4})_"
         r"v(?P<version>[\d]{3})\."
         r"(?P<type>tm)"
+    )
+
+    # Covers:
+    # MOC metakernels (type: 'mk')
+    attitude_mk_filename_pattern = (
+        r"imap_"
+        r"(?P<start_year_doy>\d{4}_\d{3})_"
+        r"a(?P<version>\d{2})\.spice\."
+        r"(?P<type>mk)"
+    )
+    ephemeris_mk_filename_pattern = (
+        r"IMAP_"
+        r"(?P<start_year_doy>\d{4}_\d{3})_"
+        r"e(?P<version>\d{2})\."
+        r"(?P<type>mk)"
     )
 
     valid_spice_regexes = (
@@ -500,6 +516,8 @@ class SPICEFilePath(ImapFilePath):
         re.compile(spice_frame_pattern),
         re.compile(sff_filename_pattern),
         re.compile(mk_filename_pattern),
+        re.compile(attitude_mk_filename_pattern),
+        re.compile(ephemeris_mk_filename_pattern),
     )
 
     class InvalidSPICEFileError(Exception):
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -368,7 +368,6 @@ class ScienceFilePath(ImapFilePath):
 _SPICE_TYPE_MAPPING = {
     "ah.bc": "attitude_history",
     "ap.bc": "attitude_predict",
-    "mk": "metakernel",
     "spin.csv": "spin",
     "repoint.csv": "repoint",
     "recon": "ephemeris_reconstructed",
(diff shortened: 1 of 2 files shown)


tests/test_file_validation.py
```diff
@@ -208,6 +208,18 @@ def test_spice_file_path():
...
     )
 
+    # MOC attitude and ephemeris metakernel files tests
+    moc_att_mk = SPICEFilePath("imap_2025_005_a01.spice.mk")
+    assert moc_att_mk.construct_path() == imap_data_access.config["DATA_DIR"] / Path(
+        "spice/mk/imap_2025_005_a01.spice.mk"
+    )
+
+    # TODO: why is tests failing here?
+    moc_ephem_mk = SPICEFilePath("IMAP_2025_005_e01.mk")
```
tech3371, line 218: This tests fails locally for me. I can't understand why. I tried to test the regex locally through python shell and it works but somehow in above code, it doesn't. I feel like I missing something small but can't see it.

imap_data_access/file_validation.py
```diff
@@ -484,14 +485,29 @@ class SPICEFilePath(ImapFilePath):
...
+        r"(?P<start_year_doy>\d{4}_\d{3})_"
+        r"a(?P<version>\d{2})\.spice\."
+        r"(?P<type>mk)"
+    )
+    ephemeris_mk_filename_pattern = (
+        r"IMAP_"
+        r"(?P<start_year_doy>\d{4}_\d{3})_"
+        r"e(?P<version>\d{2})\."
+        r"(?P<type>mk)"
+    )
```
tech3371, line 509: These two matches what's in most recent MOC data product document and I have on our [Galaxy page](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/221734667/SDC+Expected+SPICE+File+List).

tests/test_file_validation.py
tech3371: found out why. We were making it lower case before passing to match regex. I need metakernel to be case sensitive. There are so many inconsistency in MOC files that we don't have control over.

imap_data_access/file_validation.py
```diff
@@ -368,6 +368,7 @@ def is_valid_repointing(input_repointing: str) -> bool:
 _SPICE_TYPE_MAPPING = {
     "ah.bc": "attitude_history",
     "ap.bc": "attitude_predict",
+    "mk": "metakernel",
```
greglucas, line 371: Why did you choose this location? I am not sure how this list is organized now that I see this. I thought we did have other metakernels and do see a `tm` lower down. Should we put this near that entry?

imap_data_access/file_validation.py
greglucas: Can you reach out to the MOC about these inconsistencies and verify what their plan is going forward? I'd just verify this is what we are going to expect going forward or if they noticed the inconsistencies as well and were going to update them.

* The one IMAP all caps seems odd
* The optional `.spice` also seems kind of odd

imap_data_access/file_validation.py
tech3371: This dictionary is mapping spice type based on file's extension. Since mk is a metekernel type, this dictionary is saying that both file with extensions `tm` or `mk` are of type metakernel.

imap_data_access/file_validation.py
tech3371: I feel like this is beyond my control but I can ask Mac. Can we keep this for this PR and revisit later if this change?

greglucas: > I feel like this is beyond my control but I can ask Mac.

That is all I'm asking. It may not change, but it is worth making them aware.

> Can we keep this for this PR and revisit later if this change?

Yes 👍

tests/test_file_validation.py
```diff
@@ -208,6 +208,18 @@ def test_spice_file_path():
         "spice/activities/imap_0001_001_hist_00.sff"
     )
 
+    # MOC attitude and ephemeris metakernel files tests
+    moc_att_mk = SPICEFilePath("imap_2025_005_a01.spice.mk")
+    assert moc_att_mk.construct_path() == imap_data_access.config["DATA_DIR"] / Path(
+        "spice/mk/imap_2025_005_a01.spice.mk"
+    )
+
+    # TODO: why is tests failing here?
```
greglucas, line 217:
```suggestion
```

tech3371:
> Do we actually need these metakernels? I figured they might get confusing alongside the metakernels we plan on generating and using in our processing.
> 
> If we do need to use these, we should probably change the naming convention of the SDC-generated metakernels to something a bit more descriptive, like "imap_main_metakernel_YYYY_v##.tm" or "sdc_generated" or "sdc_processing" or something.

Good question. I see. That's why you didn't addd them to this. @greglucas what do you think?

greglucas:
> Good question. I see. That's why you didn't addd them to this. @greglucas what do you think?

I think we should accept them and put them somewhere at a minimum. It could be useful for debugging and seeing what metakernels they are creating versus us.
@bryan-harter makes a good point about clutter. Is the different extension enough to distinguish them? Should they be in a different directory / subfolder? Otherwise I like the short `sdc` as a descriptor to distinguish sdc versions.

bryan-harter:
Yeah we can add _sdc_ to our metakernels then. 

If we're sticking with the standard SPICE directory structure, then the "mk" folder is really the only place to put them. But I don't have a strong opinion about adhering to that structure, I'd be perfectly happy either way.

tech3371:
> Yeah we can add _sdc_ to our metakernels then.
> 
> If we're sticking with the standard SPICE directory structure, then the "mk" folder is really the only place to put them. But I don't have a strong opinion about adhering to that structure, I'd be perfectly happy either way.

Cool. Then I will change SDC maintain kernel name differently and put all of them in `spice/mk` folder.

imap_data_access/file_validation.py
tech3371: Sent email to Mac.

imap_data_access/file_validation.py
```diff
@@ -484,22 +485,39 @@ class SPICEFilePath(ImapFilePath):
     )
 
     # Covers:
-    # Metakernels (type: 'tm')
-    mk_filename_pattern = (
-        r"(imap)_"
+    # SDC generated metakernels (type: 'tm')
+    sdc_mk_filename_pattern = (
+        r"(imap)_sdc_metakernel_"
```
greglucas, line 490: I just looked at the below and we only have a start_year for our metakernels whereas the MOC ones have year/doy. Do we want ours to be year/doy too so we don't have a full year's worth of attitude needed? ping @bryan-harter 

The other benefit there is that the names are more consistent with one another. I'm guessing this will all potentially still be changing though once we get more experience with how we want to use them, so we should probably be open to changing ours later as well.

greglucas (approved):

imap_data_access/file_validation.py
bryan-harter: So my plan with these was indeed to have a full-year metakernel that contained the valid files for the whole year. At some point it sounds like ya'll wanted an on-the-fly metakernel generation API like we have on EMM, where you can specify the exact dates you need and whether or not to include things like predicts. So its possible these sdc-generated metakernel files might not even be necessary for processing. They could be a potential option until that API is in place, however. 

Additionally, I know that other missions do have these yearly metakernels (for example, here is MAVEN: https://naif.jpl.nasa.gov/pub/naif/pds/pds4/maven/maven_spice/spice_kernels/mk/). And maybe having a tangible file might be helpful for users of the data? I also think that furnishing SPICE files, even a years worth of data, is pretty low overhead. If I recall correctly, furnishing just essentially points the SPICE software to the file so it can be read in if needed.

PR #169, opened 2025-04-08
ENH: Add quicklook filepath

greglucas:
# Change Summary

## Overview

This follows the same naming convention as Ancillary Files, but a different set of extensions and download location.

One thing of note is that we have `imap/instrument` for science files, `imap/ancillary/instrument` for ancillary files, and `spice/` for spice files (notice the lack of `imap/`). Do we want to bring ancillary/quicklook up a level to be more akin to spice? Bring spice into the `imap/` directory structure? Or it is all OK and not a big deal.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import re
+import typing
 import warnings
 from abc import abstractmethod
 from datetime import datetime
@@ -15,8 +16,8 @@ import imap_data_access
 def generate_imap_file_path(filename: str) -> ImapFilePath:
     """Generate an ImapFilePath object from a filename.
 
-    This method determines if the filename is a SPICE, Science, or Ancillary file and
-    returns a SPICEFilePath, ScienceFilePath, or AncillaryFilePath object respectively.
+    This method determines if the filename is a SPICE, Science, Ancillary, or
+    Quicklook file and returns the corresponding class.
 
     Parameters
     ----------
@@ -27,26 +28,18 @@ def generate_imap_file_path(filename: str) -> ImapFilePath:
     -------
     A FilePath object
     """
-    try:
-        # SPICE
-        path_obj = imap_data_access.SPICEFilePath(filename)
-    except SPICEFilePath.InvalidSPICEFileError:
-        # Science and Ancillary
+    for cls in (ScienceFilePath, AncillaryFilePath, SPICEFilePath, QuicklookFilePath):
         try:
-            path_obj = imap_data_access.ScienceFilePath(filename)
-        except ScienceFilePath.InvalidScienceFileError:
-            # If Science file fails, then process as an Ancillary file
-            try:
-                path_obj = imap_data_access.AncillaryFilePath(filename)
-            except AncillaryFilePath.InvalidAncillaryFileError as e:
-                # Matches neither file format
-                error_message = (
-                    f"Invalid file type for {filename}. It does not match"
-                    f"Spice, Science or Ancillary file formats"
-                )
-                raise ValueError(error_message) from e
-
-    return path_obj
+            return cls(filename)
+        except (
+            ScienceFilePath.InvalidScienceFileError,
+            AncillaryFilePath.InvalidAncillaryFileError,
+            SPICEFilePath.InvalidSPICEFileError,
+        ):
+            continue
+    raise ValueError(
+        f"Invalid file type for {filename}. It does not matchany file formats."
+    )
 
 
 class ImapFilePath:
@@ -123,6 +116,9 @@ class ImapFilePath:
 class ScienceFilePath(ImapFilePath):
     """Class for building and validating filepaths for science files."""
 
+    VALID_EXTENSIONS: typing.ClassVar[set[str]] = {"cdf", "pkts"}
+    _dir_prefix = "imap"
+
     class InvalidScienceFileError(Exception):
         """Indicates a bad file type."""
 
@@ -186,6 +182,7 @@ class ScienceFilePath(ImapFilePath):
         descriptor: str,
         start_time: str,
         version: str,
+        extension: str = "cdf",
         repointing: int | str | None = None,
     ) -> ScienceFilePath:
         """Generate a filename from given inputs and return a ScienceFilePath instance.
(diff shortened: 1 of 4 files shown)


imap_data_access/file_validation.py
```diff
@@ -869,3 +869,232 @@ def extract_filename_components(filename: str | Path) -> dict:
...
+
+    def __init__(self, filename: str | Path):
+        """Class to store filepath and file management methods for Quicklook files.
+
+        If you have an instance of this class, you can be confident you have a valid
+        quicklook file and generate paths in the correct format. The parent of the file
+        path is set by the "IMAP_DATA_DIR" environment variable, defaults to "data/"
+
+        Current filename convention:
+        "<mission>_<instrument>_<descriptor>_<start_date>(_<end_date>)_
```
maxinelasp, line 890: Any reason we can't use AncillaryFilePath for this? Or subclass it since they have the same filenames and only overwrite `construct_file_path` to get the quicklook file path? (or we could update AncillaryFilePath.construct_file_path to include an optional keyword `use_quicklook` or something like that). I think a quicklook is a subset of an ancillary file, so that seems reasonable to me, and it would reduce this code.

maxinelasp (commented):
We should probably match the structure. I would probably vote for moving spice under `imap/` personally.

imap_data_access/file_validation.py
greglucas: I don't love our current setup honestly, so I agree with the sentiment here. Unfortunately, there is a lot to override surprisingly because of the extensions / validation as well. If we are OK with adding extensions to ancillary and merging these and ignore the different extensions then we can do that. I don't have a strong preference here.

maxinelasp: I think .png is fine to add as a valid AncillaryFilePath extension. Like I said, I would think of quicklooks as a type of ancillary file anyway. 

I think any case where we're trying to match on a filepath using the ImapFilePath, we would actually want to treat this as an AncillaryFilePath anyway. Maybe we should just move the quicklook plots under the ancillary folder...? that would basically entirely solve the issue, lol. Or, we could add a check for an extension in ImapFilePath (match on AncillaryFilePath, then check if the extension is in VALID_QUICKLOOK_EXTENSIONS, then return it as a QuicklookFilePath) 

I'd probably vote to just check the extension in ImapFilePath. I think any of these will work (including what you have), just a question of philosophy and maintainability

maxinelasp: The only place where I think we would have the issue a file arriving and not knowing if it's a quicklook or a "normal" ancillary would be in the indexer lambda anyway, and for those purposes treating it like a normal ancillary file would work fine... or maybe we want to exclude these from the file tracker db? I am not totally sure.

greglucas: I was viewing ancillary as "given to us", quicklooks are "produced by us". I was thinking it would be easier to maintain by keeping them separate from each other to not pollute the inputs with the outputs.

I'm happy to add the extension check into the "Ancillary" filetype and reroute based on `png`/image-type or not if others are OK with that.

maxinelasp: Ah, I have had a few requests to produce ancillary outputs...

greglucas: What would those be? I feel ancillary is such a broad catch-all term (hence your rightful justification that quicklooks are ancillary) I just may not be following everything it is being used for. I think we should be in the business of creating data products and plots and not get ourselves too extended into making other things like calibration files etc. The data product could be a "team only" product "like a quick-look but cdf/csv data" or something, but I wasn't envisioning those as ancillary.

maxinelasp: Yes, things like an output file for the team with the calculated MAG offsets during L1D. This would be a team-only thing, but I was picturing it as an ancillary file and falling into that framework.

maxinelasp: More like a processing checkpoint file than a quicklook, but a similar kind of thing

greglucas:
@maxinelasp, I have now updated this to subclass `AncillaryFilePath` now that we've decided upon the directory scheme to all be within `imap/subdir/`. This makes the class definition trivial with some minor modifications to our current AncillaryFilePath. There are still some subtle differences like it will raise an InvalidAncillaryFilePath Error etc. But routing all of that through for a new exception didn't seem worth it to me 🤷 Another issue is with the staticmethod to handle extension validation. I decided to just inject both extensions into it because the extensions are tied to the class now and didn't think it was worthwhile overriding this in both places just for that capability again. This does broaden the acceptance for both ancillary and quicklook, but that seems like a minor price to pay.

A few new concerns/thoughts that came up as I was thinking about this:
* All ancillary and quicklook files are in a flat directory structure per instrument. This could grow pretty significantly if we are making quicklook/extra files on a daily cadence (which I expect to be happening).
  * Do we want to separate this out per data level and/or dates if available?
* Do we want to allow for "repointXXX" in the ancillary/quicklook products? I am pretty sure we'll want to add some plots for a science file called `imap_hi_pset_20100101-repoint00001.cdf` and have that tied to repointing and not necessarily start/end date. So now that I'm writing this out, would we rather have `QuicklookFilePath` subclass `ScienceFilePath`?

maxinelasp (approved):
Looks great, I like your solution a lot!

For your next questions: 
```
   * All ancillary and quicklook files are in a flat directory structure per instrument. This could grow pretty significantly if we are making quicklook/extra files on a daily cadence (which I expect to be happening).
        * Do we want to separate this out per data level and/or dates if available?
        * Do we want to allow for "repointXXX" in the ancillary/quicklook products? I am pretty sure we'll want to add some plots for a science file called imap_hi_pset_20100101-repoint00001.cdf and have that tied to repointing and not necessarily start/end date. So now that I'm writing this out, would we rather have QuicklookFilePath subclass ScienceFilePath?
```

I haven't thought about this much, but initially, yes, it would be nice to add another folder for date if it exists.

I think adding repointing to quicklooks is interesting - I'm not sure if anyone needs that functionality. However, if we do, I would be ok with adding it to either AncillaryFilePath (if any ancillary files are organized on repointing, which I don't know of any) or QuicklookFilePath. As of right now, we don't have repointing in our ancillary file naming convention and I would hesitate to add it unless someone really needs it. I think from a dependency POV managing it entirely from start date is easier and makes more sense.

greglucas:
The ENA teams said that they do want the repointing option for their quicklooks, so I've now changed Quicklook to inherit from ScienceFileType instead. This required some more refactoring and breaking things out to be more generic, but all seemed reasonable to me still.

tech3371 (approved):
LGTM

PR #175, opened 2025-04-11
FIX: Repointing column was renamed to remove "time"

greglucas:
# Change Summary

## Overview

In the title. The MOC changed the repointing column name in the last update, so we failed on the last download because of this rename.


tech3371 (approved):
good catch!

PR #180, opened 2025-04-15
ENH: Switch over to requests for I/O

### LINKED ISSUE #106: Use requests library instead of urllib
`imap-data-access` currently uses `urllib` to handle GET and PUT requests, but this is causing some [known issues](https://github.com/IMAP-Science-Operations-Center/imap-data-access?tab=readme-ov-file#network-issues). It would perhaps be better to use the `requests` library instead (i.e. `requests.get()` and `requests.put()`).

greglucas:
# Change Summary

## Overview

External users have issues with certificate verification, so to make things easier on them we can use the requests library instead of the standard library. This also simplifies some of our handling of redirects and query parameter encoding as those come along for free with requests.

I've tested the download/upload/query/webpoda cli and all appear to be working as I'd expect still.

closes #106

## New Dependencies

requests

## Testing

Note that I spent some time trying to figure out how to mock a 307 response to follow a redirect, but since requests automatically handles that for us I couldn't get something to work that wasn't just testing my mock implementation. Happy for suggestions if there are any, but otherwise our critical piece is just that we don't raise on a 307 which we did previously and we don't here in the test either.

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -1,13 +1,11 @@
 """Input/output capabilities for the IMAP data processing pipeline."""
 
 import contextlib
-import json
 import logging
-import urllib.request
 from pathlib import Path
 from typing import Optional, Union
-from urllib.error import HTTPError, URLError
-from urllib.parse import urlencode
+
+import requests
 
 import imap_data_access
 from imap_data_access import file_validation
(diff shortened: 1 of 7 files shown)


tech3371 (approved):
Thank you!

PR #182, opened 2025-04-15
MNT: Improve error messages and add more logging

greglucas:
# Change Summary

## Overview

Previously, a FileNotFoundError would just display the path to the missing file as a response, but wouldn't tell you that there was an error.

Additionally, go through the code and upgrade some logs to info and add some more debug logs. This was purely subjective, so happy to change some of these if people want. The biggest thing is to add "success" INFO logs instead of just debug as people want to know when something succeeded too not just failed.

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -439,7 +439,7 @@ def main():  # noqa: PLR0915
         args.func(args)
     except Exception as e:
         # Make sure we are exiting with non-zero exit code and printing the message
-        parser.exit(status=1, message=str(e) + "\n")
+        parser.exit(status=1, message=f"{e!r}\n")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


imap_data_access/cli.py
```diff
@@ -439,7 +439,7 @@ def main():  # noqa: PLR0915
         args.func(args)
     except Exception as e:
         # Make sure we are exiting with non-zero exit code and printing the message
-        parser.exit(status=1, message=str(e) + "\n")
+        parser.exit(status=1, message=f"{e!r}\n")
```
greglucas, line 442: This gets the `__repr__` of the error message, which means we now get `FileNotFoundError(path/to/file)` instead of just `path/to/file`

sdhoyt: Nice, thank you!

sdhoyt (approved):

PR #186, opened 2025-04-23
Correct IDEX APIDs

lacoak21:
# Change Summary

## Overview
Fix idex APIDs.

Add APID 1419. This is an event catalog packet that the IDEX team wants decommed. 

APID 1418 (event message packet) will not be included because it does not have a 1 to 1 relationship with the science packets. We were originally hoping to decom it and pass it in as an ancillary file for l2b but since we dont know how to match up those packets with the science packets we are going back to the original plan of querying webpoda in l2b. 

## Updated Files
-imap_data_access/webpoda.py
   - Update apids


greglucas:
> APID 1418 (event message packet) will not be included because it does not have a 1 to 1 relationship with the science packets. We were originally hoping to decom it and pass it in as an ancillary file for l2b but since we dont know how to match up those packets with the science packets we are going back to the original plan of querying webpoda in l2b.

When you say "querying webpoda in l2b", does that mean that the l2 processing is dependent on an internet connection / availability to LASP? If that is the case I think we should get it here and it could be a separate descriptor/data product you make. It is OK that it isn't a 1-1 mapping, but making sure that file is created earlier on would be beneficial I think.

lacoak21:
The plan is now to decom 1418 and add a dependency to l2b to collect the last 7 days of event log l1a files.

greglucas (approved):

tech3371 (approved):
@lacoak21 Do you want to include in the release `v0.21.0`? Just saw that through email. I am planning to make a release ASAP and wondering if you like to have that in there.

lacoak21:
> @lacoak21 Do you want to include in the release `v0.21.0`? Just saw that through email. I am planning to make a release ASAP and wondering if you like to have that in there.

Sure thanks Tenzin! I will merge this now.

PR #189, opened 2025-04-24
Fixing repoint keyword in query

maxinelasp:
# Change Summary
This makes it so you can query against repointing using `--repointing repoint00001`. The actual API still takes in an integer, because the file tracking table uses ints instead of the `repointxxxxx` standard name. So this takes in `repoint00001` and transforms it before querying the HTTPS API. 

Before, the query failed because some parts were validating against `repointxxxxx` and some were validating against an integer. I think we should leave the repointing as an int in the database and in the query API, but match the standard on the outward facing wrapper here.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -204,7 +204,7 @@ class ScienceFilePath(ImapFilePath):
         if data_level == "l0":
             extension = "pkts"
         time_field = start_time
-        if repointing:
+        if repointing is not None:
             if ScienceFilePath.is_valid_repointing(repointing):
                 time_field += f"-{repointing}"
             elif isinstance(repointing, int):
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -167,7 +167,7 @@ class ScienceFilePath(ImapFilePath):
         descriptor: str,
         start_time: str,
         version: str,
-        repointing: int | None = None,
+        repointing: int | str | None = None,
     ) -> ScienceFilePath:
         """Generate a filename from given inputs and return a ScienceFilePath instance.
 
@@ -192,7 +192,8 @@ class ScienceFilePath(ImapFilePath):
             The version of the data
         repointing : int, optional
             The repointing number for this file, optional field that
-            is not always present
+            is not always present. Should be either a string like "repointXXXXX" or an
+            integer like 12345.
 
         Returns
         -------
@@ -204,7 +205,11 @@ class ScienceFilePath(ImapFilePath):
             extension = "pkts"
         time_field = start_time
         if repointing:
-            time_field += f"-repoint{repointing:05d}"
+            if ScienceFilePath.is_valid_repointing(repointing):
+                time_field += f"-{repointing}"
+            elif isinstance(repointing, int):
+                time_field += f"-repoint{repointing:05d}"
+
         filename = (
             f"imap_{instrument}_{data_level}_{descriptor}_{time_field}_"
             f"{version}.{extension}"
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -335,7 +335,10 @@ def main():  # noqa: PLR0915
         help="End date for a range of file timestamps in YYYYMMDD format",
     )
     query_parser.add_argument(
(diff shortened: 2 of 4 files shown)


imap_data_access/file_validation.py
```diff
@@ -204,7 +205,11 @@ def generate_from_inputs(
             extension = "pkts"
         time_field = start_time
         if repointing:
```
greglucas, line 207: I realize this was already there before, but we should change it to `is not None` so we can handle 0 too.

greglucas (commented):
> but match the standard on the outward facing wrapper here.

I'm conflicted here. I feel like for versions I would want to see the prefix "v" `v001`, but for pointings (not repointing as Tim points out so we already missed the boat on that one 😂) I feel like `repoint00001` is quite verbose. Does this come into play when we are extracting the filename components too and which version does that return (str or int)? In the processing code I would think we'll be searching through the repoint table by integer, so rather than have to do all of the stripping/converting it might be nicer to switch over to an int right away and only have the `repointXXXXX` be for the filename output.

maxinelasp:
@greglucas Like I said, it has been set in different ways in different places. I think I would expect to search via `repointxxxxx` because that's what's in the filename. In processing, you can still use an integer with the `query()` call so this is purely for the end user of the CLI. I think people will generally search directly based on what's in the filename, but if you disagree I'm happy to change it everywhere. The main problem right now is that half the time we're expecting `repointxxxxx` and half the time we're expecting an integer and it's painful. I can go through and make everything match in here and then it will be easier to use. 

I could also write some kind of repoint_to_int method that gets used everywhere that takes in an int or the `repoint00000` string and returns just an int. That could also help smooth out our interfaces everywhere.

greglucas:
> The main problem right now is that half the time we're expecting repointxxxxx and half the time we're expecting an integer and it's painful.

Agreed!

> In processing, you can still use an integer with the query() call so this is purely for the end user of the CLI. I think people will generally search directly based on what's in the filename, but if you disagree I'm happy to change it everywhere.

I agree with you, and I think it is fine to allow search based on either version. I think my question is what the returns are from the various methods. Can we make the `extract_filename_components()` return a `repointing = 1` integer and then I think everything would be consistent with ints round-tripping everywhere. You could then add the query capability of `repointing00001`, but the return table would show `1`. Does this make sense?

maxinelasp:
@greglucas Yes, I understand. File validation uses integers everywhere except `is_valid_repointing()` which is used to check the output filename and so checks for the `repoint` prefix. Right now, I believe all software importing imap-data-access can use repointing as a pure integer, and these changes are only affecting the CLI, except for the change which adds functionality so users can pass in an integer OR a repointxxxxx string into generate_from_inputs, which just made the CLI updates a little cleaner and moves the verification/int stripping pieces into that function rather than in the CLI.

 I can remove that functionality if you would prefer it to be more int-focused everywhere else, as it is I like having the option but it does continue to encourage the disparity.

greglucas:
> Right now, I believe all software importing imap-data-access can use repointing as a pure integer, and these changes are only affecting the CLI, except for the change which adds functionality so users can pass in an integer OR a repointxxxxx string into generate_from_inputs, which just made the CLI updates a little cleaner and moves the verification/int stripping pieces into that function rather than in the CLI.

Thank you for taking the time to explain this to me and sorry it took me so long to pick up on this! I think I am just confused as to what was internally happening. Looking through the code, `is_valid_repointing()` is only used in the io routines not the file_validation portion. The `extract_components()` returns an integer for repointing (I did not realize this and thought it was the string version).

One minor additional point: I think this removes the ability to query by an integer repointing. You add the string/int case to the generate filenames, but the query input is a string through the cli and thus would be "1" in that case and not be validated properly. Can you add both options to that support as well? Then I think I'm on board with this update.

maxinelasp:
@greglucas Is this what you were thinking?

greglucas:
yep something like that. I wasn't sure if it should be updated in the isvalid check or like this, so looks good to me.

imap_data_access/file_validation.py
greglucas: @maxinelasp can you update this too?

PR #194, opened 2025-05-02
ENH: Adding SPICEInput capabilities

tech3371:
# Change Summary

## Overview
Adding features to SPICEInput class to take in SPICE files. This will allow SPICE to be added on batch starter and imap_processing CLI.

## Updated Files
   - Added features to support SPICE files in the ProcessingInputCollection

## Testing
   - Good amount of tests to show how SPICE will be added and queried in batch starter and `imap_processing`'s `pre_processing()`

### DIFF
--- a/imap_data_access/processing_input.py
+++ b/imap_data_access/processing_input.py
@@ -24,8 +24,6 @@ class ProcessingInputType(Enum):
     SCIENCE_FILE = "science"
     ANCILLARY_FILE = "ancillary"
     SPICE_FILE = "spice"
-    SPIN_FILE = "spin"
-    REPOINT_FILE = "repoint"
 
 
 class InputTypePathMapper(Enum):
@@ -34,6 +32,14 @@ class InputTypePathMapper(Enum):
     SCIENCE_FILE = ScienceFilePath
     ANCILLARY_FILE = AncillaryFilePath
     SPICE_FILE = SPICEFilePath
+
+
+class SPICESource(Enum):
+    """Enum matching source of SPICE file types."""
+
+    SPICE = "spice"
+    SPIN = "spin"
+    REPOINT = "repoint"
 
 
 @dataclass
@@ -279,17 +285,15 @@ class SPICEInput(ProcessingInput):
         These attributes are used to group and manage SPICE files effectively.
 
         Key Attributes:
-        1. input_type: Identifies the type of input
+        1. input_type: Identifies the type of input. ProcessingInputType.SPICE_FILE
+        2. source: Specifies the source of the files
             (e.g., 'spice', 'spin', 'repoint'). This helps in serialize() output. Eg.
            [
                 {"type": "spice", "files": [ordered list of SPICE files]},
                 {"type": "spin", "files": [<list of spin files>]},
                 {"type": "repoint", "files": [<latest repoint file>]}
            ]
-        2. source: Specifies the source of the files
-            (e.g., 'spice', 'spin', 'repoint').
-        3. data_type: Groups and read files by their type
-            (e.g., 'spice', 'spin', 'repoint').
+        3. data_type: Indicates the type of data. ProcessingInputType.SPICE_FILE.value
         4. descriptor: Indicates the file descriptor ('historical' by default, or 'best'
             if predictive kernels are included).
 
@@ -298,6 +302,8 @@ class SPICEInput(ProcessingInput):
         args : str
             Input SPICE filenames.
         """
+        self.input_type = ProcessingInputType.SPICE_FILE
+        self.data_type = ProcessingInputType.SPICE_FILE.value
         self.descriptor = "historical"
         super().__init__(*args)
 
@@ -320,29 +326,31 @@ class SPICEInput(ProcessingInput):
 
         if "spin" in source:
             # Update the source to be spin
-            self.source = "spin"
-            self.input_type = ProcessingInputType.SPIN_FILE
-            self.data_type = ProcessingInputType.SPIN_FILE.value
+            self.source = SPICESource.SPIN.value
             if len(source) != 1:
                 raise ValueError(
                     "If spin data in the list, it should only contain spin files"
                 )
         elif "repoint" in source:
             # Update the source to be repoint
-            self.source = "repoint"
-            self.input_type = ProcessingInputType.REPOINT_FILE
-            self.data_type = ProcessingInputType.REPOINT_FILE.value
+            self.source = SPICESource.REPOINT.value
             # Latest file will contain all the repointing data.
             if len(file_obj_list) != 1:
                 raise ValueError(
                     "There should only be one repoint file in the list of files"
                 )
         else:
-            self.source = ProcessingInputType.SPICE_FILE.value
-            self.input_type = ProcessingInputType.SPICE_FILE
-            self.data_type = ProcessingInputType.SPICE_FILE.value
+            self.source = SPICESource.SPICE.value
 
         self.imap_file_paths = file_obj_list
+
+    def construct_json_output(self):
+        """Construct a JSON output.
+
+        This contains the minimum information needed to construct an identical
+        ProcessingInput instance (input_type and filename)
+        """
+        return {"type": self.source, "files": self.filename_list}
 
     def get_time_range(self):
         """Not yet complete."""
@@ -451,7 +459,6 @@ class ProcessingInputCollection:
         self,
         source: str | None = None,
         descriptor: str | None = None,
-        data_type: str | None = None,
     ) -> list[Path]:
         """Get the dependency files path from the collection.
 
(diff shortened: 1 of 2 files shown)


imap_data_access/processing_input.py
```diff
@@ -271,26 +271,52 @@ class SPICEInput(ProcessingInput):
...
-        self.data_type = ProcessingInputType.SPICE_FILE.value
-        self.descriptor = "predict"
+        for file in self.filename_list:
+            path_validator = SPICEFilePath(file)
+            kernel_type = path_validator.spice_metadata["type"]
+            source.add(kernel_type)
+            file_obj_list.append(path_validator)
+
+            # Set the descriptor to be predict if it contains any predict kernel types
+            if (
```
lacoak21, line 296: So valid sources are now only ["spin", "repoint", "spice"]?

The sds-data-manager currently has: 
    SC_ATTITUDE: str = "sc_attitude"
    SC_EPHEMERIS: str = "sc_ephemeris"
    PLANET_EPHEMERIS: str = "planet_ephemeris"
    TIME_KERNEL: str = "time_kernel"
    THRUSTER_FIRE_KERNEL: str = "thruster_fire_kernel"
    SC_SPIN: str = "sc_spin"
    SC_REPOINT: str = "sc_repoint"
    SC_POINTING_FRAME: str = "sc_pointing_frame"`

imap_data_access/processing_input.py
```diff
@@ -271,26 +271,52 @@ class SPICEInput(ProcessingInput):
...
+                "ephemeris" in kernel_type and kernel_type != "ephemeris_reconstructed"
+            ) or kernel_type == "attitude_predict":
+                self.descriptor = "best"
```
lacoak21, line 299: What about "reconstruct" and "nominal" spice descriptors?

imap_data_access/processing_input.py
```diff
@@ -271,26 +271,52 @@ class SPICEInput(ProcessingInput):
...
+
+        if "spin" in source:
+            # Update the source to be spin
+            self.source = "spin"
+            if len(source) != 1:
```
lacoak21, line 304: For ancillary and science we only allow one source type in each ProcessingInput - is there a reason why we are not doing this for all sources except spin?

lacoak21 (commented):
Looks good! Had some questions about the updated sources and descriptors.

imap_data_access/processing_input.py
lacoak21: Is this so we can furnish everything at once or related to meta kernel?

imap_data_access/processing_input.py
tech3371: Good point. Valid sources are those code block still. I need to remove that class on that repo and reference kernel types defined in this repo.

I was using `["spin", "repoint", "spice"]` to figure out which data to group by and query for. This is me thinking about how we query for SPICE kernels, spin or repoint data in `pre_processing()`. I thought of putting those `["spin", "repoint", "spice"]` in `data_type` but then I need to update `get_file_paths()` in the collection class. I think that's a small price to pay...I will do that since I see your confusion.

current code passes data like this
```
[
{
    "type": 'spice',
    "files": [
        "naif0012.tls",
        "imap_sclk_0001.tsc",
        "pck00010.tpc",
        "imap_001.tf",
        "de440.bsp",
        "imap_90days_20260922_20261221_v01.bsp",
        "imap_1000_100_1000_100_01.ah.bc",
        "imap_1000_100_1000_100_02.ap.bc",
    ]
},
{
    "type": 'spice',
    "files": [
        <spin-files>
    ]
},
{
    "type": 'spice',
    "files": [
        <latest-repoint-file>
    ]
}
]
```
With `data_type` as `[spice, spin, repoint]`, then it will look like this.
```
[
{
    "type": 'spice',
    "files": [
        "naif0012.tls",
        "imap_sclk_0001.tsc",
        "pck00010.tpc",
        "imap_001.tf",
        "de440.bsp",
        "imap_90days_20260922_20261221_v01.bsp",
        "imap_1000_100_1000_100_01.ah.bc",
        "imap_1000_100_1000_100_02.ap.bc",
    ]
},
{
    "type": 'spin',
    "files": [
        <spin-files>
    ]
},
{
    "type": 'repoint',
    "files": [
        <latest-repoint-file>
    ]
}
]
```

imap_data_access/processing_input.py
tech3371: This is changing based on recent discussion. If list of kernel contains any predictive kernel, then I was thinking of using 'best'. otherwise, 'historical'.

imap_data_access/processing_input.py
tech3371: Yes. It's related to your second point. SPICE, we have to do things little bit differently. If we create one ProcessingInput for each kernel types, the dependency list can become long and hard to maintain order of kernels which is very important, I learned. That's why, we decided that all kernel types will go in one SPICE dependency. Eg.
```
[
{
    "type": 'spice',
    "files": [
        "naif0012.tls",
        "imap_sclk_0001.tsc",
        "pck00010.tpc",
        "imap_001.tf",
        "de440.bsp",
        "imap_90days_20260922_20261221_v01.bsp",
        "imap_1000_100_1000_100_01.ah.bc",
        "imap_1000_100_1000_100_02.ap.bc",
    ]
},
{
    "type": 'spice',
    "files": [
        <spin-files>
    ]
},
{
    "type": 'spice',
    "files": [
        <latest-repoint-file>
    ]
}
]
```

imap_data_access/processing_input.py
lacoak21: Ah I see now! I think that is a great solution and logically makes sense in regards to querying. If you add a comment I think that would be enough.

imap_data_access/processing_input.py
lacoak21: Seems very logical to me 👍

lacoak21 (approved):
Thanks for explaining the stuff I was behind on.

imap_data_access/processing_input.py
tech3371: cool. I made that change but will add comment. Thank you for thorough review. You connected dots and raised really good point. It's very helpful!

imap_data_access/processing_input.py
```diff
@@ -271,26 +273,76 @@ class SPICEInput(ProcessingInput):
...
         """Initialize the attributes from the SPICE file name.
 
-        Not completed.
+        The SPICEInput class customizes the initialization process to handle
+        SPICE-specific requirements. It sets attributes such as input type,
+        source, data type, and descriptor based on the provided filenames.
+        These attributes are used to group and manage SPICE files effectively.
+
+        Key Attributes:
+        1. input_type: Identifies the type of input
```
maxinelasp, line 282: You have some duplicate attributes here. `input_type`, `source` and data_type` all contain the same information. I have a proposal to fix it:

Judging from your tests, you want the output to be grouped like this: 
```
[
{
    "type": 'spice',
    "files": [
        "naif0012.tls",
        "imap_sclk_0001.tsc",
        "pck00010.tpc",
        "imap_001.tf",
        "de440.bsp",
        "imap_90days_20260922_20261221_v01.bsp",
        "imap_1000_100_1000_100_01.ah.bc",
        "imap_1000_100_1000_100_02.ap.bc",
    ]
},
{
    "type": 'spin',
    "files": [
        <spin-files>
    ]
},
{
    "type": 'repoint',
    "files": [
        <latest-repoint-file>
    ]
}
]
```

So that's what I'm aiming for here!

I think `input_type` should always be `ProcessingInputType.SPICE`, because this `SPICEInput` and `SPICEFilePath` are used for all the files, and that's kind of what `input_type` is used for - to indicate the classes used for that file. They're all grouped under the SPICE umbrella from the outside view.

Then, I think `source` should be where you define SPICE vs spin vs repoint. Maybe it would make sense to add a new enum called SpiceSource which can define the options here, or you can just leave this variable exactly as you have it.

Then, data_type and descriptor can be merged. One of them should have the file descriptor ("historical", "best") and the other can just be unused if it doesn't make sense for SPICE (just name it "spice" or something.) Right now you use `descriptor` for that, so that can also be unchanged, and then `data_type` just won't really be used here.  

To get the output you're looking for, override `construct_json_output` to use the source instead of the type when building the serialized output. 

This way, you only have the information in one place, and it lines up better with the way the variables are used in the other ProcessingInput classes. 

Thoughts?

maxinelasp (changes requested):
Sorry for the slow review! I think there are a lot of duplicate attributs here and it would be nice to consolidate them, but besides that it looks good! (I left a long comment with more info)

imap_data_access/processing_input.py
```diff
@@ -24,6 +24,8 @@ class ProcessingInputType(Enum):
     SCIENCE_FILE = "science"
     ANCILLARY_FILE = "ancillary"
     SPICE_FILE = "spice"
+    SPIN_FILE = "spin"
+    REPOINT_FILE = "repoint"
 
 
 class InputTypePathMapper(Enum):
```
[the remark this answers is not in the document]
tech3371: I see your point. I thought about the latter part but it causes to much change that's unnecessary. I have thought about where spin and repoint fits a lot and this was best approach imo. I read your suggestion below. That makes sense and will make change that kind of lines up with that suggestions and undo changes in `ProcessingInputType`.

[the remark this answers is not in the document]
maxinelasp: Sounds good! I wrote this comment before the rest of my review, so yeah some of the other changes might make this unnecessary.

maxinelasp (approved):
Nice, thanks for making those updates!

PR #197, opened 2025-05-06
ENA DPS - added dps filename convention

laspsandoval:
# Change Summary

## Overview
Add new filename convention to imap-data-access's file_validation.py file. Update _SPICE_TYPE_MAPPING and _SPICE_DIR_MAPPING. Then add regex in SPICEFilePath() class.

For filename: imap_dps_yyyy_doy-repoint<xxxxx>_##.ah.bc

## Updated Files
- file_validation.py
   - added dps filename convention

## Testing
- test_file_validation.py

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -409,7 +409,7 @@ _SPICE_TYPE_MAPPING = {
     "de": "planetary_ephemeris",
     "pck": "planetary_constants",
     "naif": "leapseconds",
-    "imap_dps": "dps_history",
+    "imap_dps": "pointing_attitude",
     "imap_sclk_": "spacecraft_clock",
     "tf": "frames",
     "mk": "metakernel",
@@ -419,7 +419,7 @@ _SPICE_TYPE_MAPPING = {
 
 _SPICE_DIR_MAPPING = {
     "attitude_history": "ck",
-    "dps_history": "ck",
+    "pointing_attitude": "ck",
     "attitude_predict": "ck",
     "spin": "spin",
     "repoint": "repoint",
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -472,7 +472,7 @@ class SPICEFilePath(ImapFilePath):
     # DPS kernel (type: ah.bc)
     dps_file_pattern = (
         r"(?P<type>imap_dps)_"
-        r"(?P<start_year_doy>\d{4}_\d{3})_"
+        r"(?P<start_year_doy>\d{4}_\d{3})-"
         r"repoint(?P<repointing>\d{5})_"
         r"(?P<version>\d+)\."
         r"(?P<extension>ah\.bc)"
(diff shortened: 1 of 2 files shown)


laspsandoval:
pre-commit.ci autofix

imap_data_access/file_validation.py
```diff
@@ -409,6 +409,7 @@ def is_valid_for_start_date(self, start_date: datetime) -> bool:
     "de": "planetary_ephemeris",
     "pck": "planetary_constants",
     "naif": "leapseconds",
+    "imap_dps": "dps_history",
```
subagonsouth, line 412: I'm conflicted about calling this the `dps_history` file. I still like using `dps` in the filename since having `pointing` in there would be confusing with the `repointNNNN` field. But I do wonder if some places, it should be referred to as the `pointing_attitude`?

tech3371: yeah, at our team tag-up, looks like we want to go away from `dps` nomenclature to reduce confusion. Just wanted to mention that. But I am happy with either names that you both are happy with.

Another thing I want to mention for awareness is that this line is saying that if there is `imap_dps` in any filename, it will consider that as `dps_history` SPICE file type. Then in dependency tree, someone will be using that to state that, they need `dps_history, spice, historical, ultra, l1c, pset, UPSTREAM, HARD`. Just want to mention that to connect where this will used down the pipeline of this.

tests/test_file_validation.py
```diff
@@ -223,6 +223,25 @@ def test_spice_file_path():
         "imap/spice/mk/IMAP_2025_005_e01.mk"
     )
 
+    dps_pointing_file = SPICEFilePath("imap_dps_2025_121-repoint00023_007.ah.bc")
+    assert dps_pointing_file.construct_path() == imap_data_access.config[
+        "DATA_DIR"
+    ] / Path("imap/spice/ck/imap_dps_2025_121-repoint00023_007.ah.bc")
```
tech3371, line 229: It's good to see that it goes into attitude kernel folder.

tech3371 (commented):
I am going to ask @subagonsouth to approve this since this filename convention and what SPICE fie type it is will be impacting ENA and you guys know best. The code changes itself looks great. It's the philosophical logic that I am unsure about.

imap_data_access/file_validation.py
subagonsouth: Sorry, @laspsandoval. This is my fault.

The name of the SPICE frame is `IMAP_DPS`. I think that we are too far past that decision to change that.

So, does it make sense to have the filename be `imap_dps_yyyydoy_repointNNNNN_##.ah.bc` and then refer to the kernel as the pointing kernel? Or should we replace `dps` in the filename with `pointing` or `pf` (pointing frame)?

laspsandoval: Personally I like what we decided on:
imap_dps_yyyydoy_repointNNNNN_##.ah.bc

I could change the above dps_history to just dps.

subagonsouth (approved):
LGTM. Thanks for doing this!

laspsandoval:
pre-commit.ci autofix

PR #199, opened 2025-05-07
Add a new filtering method to ProcessingInputCollection

maxinelasp:
# Change Summary
I wanted to filter out ancillary fileinputs but there wasn't a method to do so. So this is a general filtering method for ProcessingInputCollection. 

It does replace get_science_inputs() but that's used elsewhere so I just wrapped the new method. 

## Overview

- Add new get_processing_inputs method
- Update existing methods to use get_processing_inputs internally

### DIFF
--- a/imap_data_access/processing_input.py
+++ b/imap_data_access/processing_input.py
@@ -375,6 +375,9 @@ class ProcessingInputCollection:
     def get_science_inputs(self, source: str | None = None) -> list[ProcessingInput]:
         """Return just the science files from the collection.
 
+        NOTE: get_processing_inputs is a more general method that should be used instead
+        of this one.
+
         Parameters
         ----------
         source : str, optional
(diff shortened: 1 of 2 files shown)


imap_data_access/processing_input.py
```diff
@@ -435,6 +435,9 @@ def deserialize(self, json_input: str) -> None:
     def get_science_inputs(self, source: str | None = None) -> list[ProcessingInput]:
         """Return just the science files from the collection.
 
+        NOTE: get_processing_inputs is a more general method that should be used instead
```
lacoak21, line 438: I can switch over to using get_processing_inputs if you want to remove this. Although I do not know if this is used elsewhere.

lacoak21 (approved):
Looks great - this is a better solution for sure

imap_data_access/processing_input.py
maxinelasp: It is not. I would appreciate you switching over, but just try and remember next time you're editing the code. Not a big deal right now!

PR #202, opened 2025-05-07
Add return latest bool for processingInputCollection filtering method

lacoak21:
# Change Summary

## Overview
Add handling to function "get_valid_inputs_for_start_date" to only return one file per AncillaryInput with the latest start date. 

## Updated Files

    - return latest file
    - This is only ancillary because there should be only one file valid for a date for science and SPICE should not be filtering files valid for that date.

## Testing
add checks to make sure only the latest file is getting returned when specified.


maxinelasp (approved):
Nice, looks good

lacoak21:
pre-commit.ci autofix

PR #205, opened 2025-05-13
Fixing Ingestion CLI Type Error

daralynnrhode:
# Change Summary

## New Files

- new file 1
   - description of new file 1's purpose

## Deleted Files

- deleted file 1
   - explanation for why file was deleted

## Updated Files

- updated file 1
   - description of change 1 in file 1
   - description of change 2 in file 2
- updated file 2
   - description of change 1 in file 2

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -51,7 +51,7 @@ def _print_query_results_table(query_results: list[dict]):
     if num_files == 0:
         return
 
-    # Use the keys of the first item in query_results for the header
+    # Use the query_results for the header
     headers = [
         "Instrument",
         "Data Level",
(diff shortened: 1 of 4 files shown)


daralynnrhode:
pre-commit.ci autofix

imap_data_access/cli.py
```diff
@@ -64,13 +64,21 @@ def _print_query_results_table(query_results: list[dict]):
...
         )
-        # Calculate the maximum width for each column based on the header and the data
+
+        column_widths["Ingestion Date"] = max(
+            len("Ingestion Date"),
+            *(
+                len(os.path.basename(item.get("ingestion_date", "")))
+                for item in query_results
+            ),
+        )
```
bourque, line 81: Do we need to treat Ingestion Date separately here, or could that just be taken care of by lines 69-73? I would think the header of "Ingestion Date" will always be longer than the values of the results since the header should always be more characters than what is in "`YYYYMMDD`"

daralynnrhode: I believe it does have to be treated separately, since the output of the ingestion date column ends up being longer than the actual column header (date and time are both printed). I'll attach an image of how the header row got misaligned. 
Screen Shot 2025-05-13 at 10 23 03 AM

bourque: Ah ok, I wasn't considering the time component.

bourque (approved):
Looks good!

maxinelasp (approved):

PR #208, opened 2025-05-14
Redoing SPICEInput parameters to solve inconsistency in implementation

### LINKED ISSUE #207: MNT: Redoing SPICEInput Parameters
This is result from SPICE implementation. This helps simplify some of logics in batch starter.

This adds feature to get back input object based on inputs file. This is similar to what we are doing with validating filename convention.

Updated source and data_type of SPICEInput to be able to line up with what batch starter needs and it also lines up nicely with metakernel API requests.

tech3371:
# Change Summary
closes #207 

## Overview
Adding changes to simplify batch starter logics.

## Updated Files
- imap_data_access/__init__.py
   - adding new classes for easier import
   - main changes related to what batch starter need.

## Testing

### DIFF
--- a/imap_data_access/processing_input.py
+++ b/imap_data_access/processing_input.py
@@ -33,32 +33,18 @@ def generate_imap_input(filename: str) -> ProcessingInput:
     -------
     A FilePath object
     """
-    try:
-        # SPICE
-        path_obj = SPICEInput(filename)
-    except (SPICEInput.InvalidSPICEInputError, SPICEFilePath.InvalidSPICEFileError):
-        # Science and Ancillary
+    for cls in (ScienceInput, AncillaryInput, SPICEInput, SpinInput, RepointInput):
         try:
-            path_obj = ScienceInput(filename)
+            return cls(filename)
         except (
-            ScienceInput.InvalidScienceInputError,
             ScienceFilePath.InvalidScienceFileError,
+            AncillaryFilePath.InvalidAncillaryFileError,
+            SPICEFilePath.InvalidSPICEFileError,
         ):
-            # If Science file fails, then process as an Ancillary file
-            try:
-                path_obj = AncillaryInput(filename)
-            except (
-                AncillaryInput.InvalidAncillaryInputError,
-                AncillaryFilePath.InvalidAncillaryFileError,
-            ) as e:
-                # Matches neither file format
-                error_message = (
-                    f"Invalid input type for {filename}. It does not match"
-                    f"Spice, Science or Ancillary file formats"
-                )
-                raise ValueError(error_message) from e
-
-    return path_obj
+            continue
+    raise ValueError(
+        f"Invalid input type for {filename}. It does not match any file formats."
+    )
 
 
 class ProcessingInputType(Enum):
@@ -122,6 +108,11 @@ class ProcessingInput(ABC):
     source: str = field(init=False)
     data_type: str = field(init=False)  # should be data level or "ancillary" or "spice"
     descriptor: str = field(init=False)
+
+    class ProcessingInputError(Exception):
+        """Indicate that the ProcessingInput is invalid."""
+
+        pass
 
     def __init__(self, *args):
         """Initialize using a list of filepaths and sets the attributes of the class.
@@ -259,11 +250,6 @@ class AncillaryInput(ProcessingInput):
     and descriptor.
     """
 
-    class InvalidAncillaryInputError(Exception):
-        """Indicate that the AncillaryInput is invalid."""
-
-        pass
-
     # Can contain multiple ancillary files - should have the same descriptor
     def __init__(self, *args):
         """Set the processing type to AncillaryFile and then calls super().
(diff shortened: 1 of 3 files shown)


imap_data_access/processing_input.py
```diff
@@ -18,6 +18,49 @@
 from imap_data_access.io import download
 
 
+def generate_imap_input(filename: str) -> ProcessingInput:
```
greglucas, line 21: I am working on refactoring the file_path version of this to accept more options and have the same suggestion here. We should loop over the options IMO rather than keep nesting try/excepts. Here is the code copied over since I haven't pushed that yet.

```
def generate_imap_file_path(filename: str) -> ImapFilePath:
    """Generate an ImapFilePath object from a filename.

    This method determines if the filename is a SPICE, Science, Ancillary, or
    Quicklook file and returns the corresponding class.

    Parameters
    ----------
    filename : str
        The filename to generate a path for.

    Returns
    -------
    A FilePath object
    """
    for cls in (ScienceFilePath, AncillaryFilePath, SPICEFilePath, QuicklookFilePath):
        try:
            return cls(filename)
        except (
            ScienceFilePath.InvalidScienceFileError,
            AncillaryFilePath.InvalidAncillaryFileError,
            SPICEFilePath.InvalidSPICEFileError,
        ):
            continue
    raise ValueError(
        f"Invalid file type for {filename}. It does not matchany file formats."
    )
```

imap_data_access/processing_input.py
```diff
@@ -168,6 +211,11 @@ class ScienceInput(ProcessingInput):
      and descriptor.
     """
 
+    class InvalidScienceInputError(Exception):
```
greglucas, line 214: Personally, I think this is too many errors, can we just have a single `ProcessingInputError` that we catch everywhere rather than needing to catch all of these separately? The message to each when we raise can be different if we want.

imap_data_access/processing_input.py
```diff
@@ -325,22 +386,27 @@ def _set_attributes_from_filenames(self) -> None:
...
             # Latest file will contain all the repointing data.
             if len(file_obj_list) != 1:
                 raise ValueError(
                     "There should only be one repoint file in the list of files"
                 )
         else:
-            self.source = SPICESource.SPICE.value
+            # Update the source and data type to be file type and spice, respectively
+            # Join all the sources into a single string separated by commas
+            self.source = ",".join(source)
```
greglucas, line 408: Why do we convert these to strings, could we keep it as a list instead? I feel like the to/from string should be done on casting to the call sites that need strings, but otherwise it seems nicer as a list of items internally.

imap_data_access/processing_input.py
tech3371: This is much better!

imap_data_access/processing_input.py
```diff
@@ -18,6 +18,36 @@
...
+    -------
+    A FilePath object
+    """
+    for cls in (ScienceInput, AncillaryInput, SPICEInput, SpinInput, RepointInput):
+        try:
+            return cls(filename)
+        except (
+            ScienceFilePath.InvalidScienceFileError,
+            AncillaryFilePath.InvalidAncillaryFileError,
+            SPICEFilePath.InvalidSPICEFileError,
```
greglucas, line 42: Do you need the Path errors here too, or can you just catch the `ProcessingInputError`?

imap_data_access/processing_input.py
```diff
@@ -274,83 +309,104 @@ def get_file_for_time(self, day):
 
 
 class SPICEInput(ProcessingInput):
-    """SPICE file subclass for ProcessingInput."""
+    """SPICE kernel file subclass for ProcessingInput."""
 
     def __init__(self, *args) -> None:
-        """Initialize the attributes from the SPICE file name.
```
greglucas, line 280: Do you want to keep all of this documentation around somewhere?

imap_data_access/processing_input.py
```diff
@@ -274,83 +309,104 @@ def get_file_for_time(self, day):
...
+
+class RepointInput(ProcessingInput):
+    """Repoint file subclass for ProcessingInput."""
+
+    def __init__(self, *args) -> None:
+        """Initialize the attributes for repoint files."""
+        self.input_type = ProcessingInputType.SPICE_FILE
+        self.source = SPICESource.REPOINT.value
+        self.data_type = SPICESource.REPOINT.value
+        self.descriptor = "historical"
```
greglucas, line 396: On all of these classes I think these could be class attributes and you could just inherit the __init__ without needing to do anything here.

imap_data_access/processing_input.py
```diff
@@ -274,83 +309,104 @@ def get_file_for_time(self, day):
...
+        pass
+
+
+class RepointInput(ProcessingInput):
+    """Repoint file subclass for ProcessingInput."""
+
+    def __init__(self, *args) -> None:
+        """Initialize the attributes for repoint files."""
+        self.input_type = ProcessingInputType.SPICE_FILE
+        self.source = SPICESource.REPOINT.value
```
greglucas, line 394: I don't think I fully understood this point when we were discussing earlier. This is still an input_type of SPICE_FILE, just the source and data_type have changed. For some reason I was thinking the input_type was also going to be Repoint/Spin too.

Just writing this down in case this is something missed here.

greglucas (approved):
I think this looks good based on our discussion earlier today. Just some minor suggestions for code style.

imap_data_access/processing_input.py
tech3371: we need it. I tried it before without those but it would not get to processing error if file validation doesn't pass first.

imap_data_access/processing_input.py
tech3371: I understand. There are so many underlying logic that hard to capture. Yeah, we need to keep all three of them under `input_type='spice'` because that's what's used to decide which validation validation function to use. All three of them of uses same file validation class. We are creating different classes here to line up with number of data_type like you mentioned yesterday. At least, that was my understanding.

To your previous comment, I will bring back those doc string comment and capture that in these classes.

imap_data_access/processing_input.py
tech3371: I am hesitant to do that because Processing input is inherit by both science and ancillary and initializing it to `historical` can cause confusion. If you are thinking about these two classes inherit from `SPICEInput()`, that I can do. Is that what you are thinking?

greglucas (approved):

PR #222, opened 2025-05-21
Added Science Frame Regex

### LINKED ISSUE #221: Add Science frame kernel to file validation
The science frame kernel (ie imap_science_0001.tf) needs to be added to the file validation in order to upload to the SDC.

sdhoyt:
# Change Summary

## Overview
This PR adds the regex for the SPICE science frame to the file validation regex

## Updated Files
- file_validation
   - added regex for science frame

## Testing
- added `test_spice_extract_science_frame_parts` test
- added assert to `test_spice_file_path`

Closes #221

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -496,7 +496,6 @@ class SPICEFilePath(ImapFilePath):
         r"(?P<version>\d+)\."
         r"(?P<extension>ah\.bc)"
     )
-
     # Covers:
     # Repoint Files (type: repoint.csv)
     repoint_file_pattern = (
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -428,7 +428,8 @@ _SPICE_TYPE_MAPPING = {
     "naif": "leapseconds",
     "imap_dps": "pointing_attitude",
     "imap_sclk_": "spacecraft_clock",
-    "tf": "frames",
+    "tf": "imap_frames",
+    "imap_science": "science_frames",
     "mk": "metakernel",
     "tm": "metakernel",
     "sff": "thruster",
@@ -450,7 +451,8 @@ _SPICE_DIR_MAPPING = {
     "planetary_constants": "pck",
     "leapseconds": "lsk",
     "spacecraft_clock": "sclk",
-    "frames": "fk",
+    "imap_frames": "fk",
+    "science_frames": "fk",
     "metakernel": "mk",
     "thruster": "activities",
 }
@@ -532,11 +534,11 @@ class SPICEFilePath(ImapFilePath):
 
     # Covers:
     # Frame: (type: 'tf')
-    spice_frame_pattern = r"(imap)_(?P<version>[\d]+)\.(?P<type>tf)"
+    imap_frame_pattern = r"(imap)_(?P<version>[\d]+)\.(?P<type>tf)"
 
     # Covers:
-    # Science Frame: (type: 'tf')
-    science_frame_pattern = r"(imap_science)_(?P<version>[\d]+)\.(?P<type>tf)"
+    # Science Frame: (type: 'imap_science')
+    science_frame_pattern = r"(?P<type>imap_science)_(?P<version>[\d]+)\.tf"
 
     # Covers:
     # Thruster files (type: sff)
@@ -580,7 +582,7 @@ class SPICEFilePath(ImapFilePath):
         re.compile(repoint_file_pattern, re.IGNORECASE),
         re.compile(spacecraft_ephemeris_file_pattern, re.IGNORECASE),
         re.compile(spice_prod_ver_pattern, re.IGNORECASE),
-        re.compile(spice_frame_pattern, re.IGNORECASE),
+        re.compile(imap_frame_pattern, re.IGNORECASE),
         re.compile(sff_filename_pattern, re.IGNORECASE),
         re.compile(sdc_mk_filename_pattern),
         re.compile(attitude_mk_filename_pattern, re.IGNORECASE),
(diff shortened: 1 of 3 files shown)


sdhoyt:
pre-commit.ci autofix

tests/test_file_validation.py
```diff
@@ -345,6 +350,17 @@ def test_spice_extract_repoint_parts():
     assert len(file_path.spice_metadata) == 5
 
 
+def test_spice_extract_science_frame_parts():
+    file_path = SPICEFilePath("imap_science_0001.tf")
+    assert file_path.spice_metadata["version"] == "0001"
+    assert file_path.spice_metadata["type"] == "frames"
```
subagonsouth, line 356: I think that this needs to be updated to `"science_frames"` and the other frames kernel should be `"imap_frames"`. See: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/228069149/SDC+SPICE+Dependency+Association+SPIKE#SDCSPICEDependencyAssociationSPIKE-SupportedKernelTypes

@tech3371, do you agree?

subagonsouth (commented):
Thanks for doing this @sdhoyt! My only concern is that we need to distinguish between the imap spacecraft frame kernel (I have called this the instrument frame kernel) and the science frame kernel.

tests/test_file_validation.py
tech3371: In above `science_frame_pattern`, it's done like this `.(?P<type>tf)` and that means that it's of "type" `tf` and then `tf` is mapped to as `frames` type. So syntax wise, it looks correct.

If you guys like that to be of type `science_frames`, then we would need to define a new mapping in `_SPICE_TYPE_MAPPING` and `_SPICE_DIR_MAPPING` where in first dictionary, you would defined what the type it is from regex and that needs to map to certain type and in second you map that certain type into correct folder path. I will make some suggestions changes if that's what you guys want.

imap_data_access/file_validation.py
```diff
@@ -534,6 +534,10 @@ class SPICEFilePath(ImapFilePath):
     # Frame: (type: 'tf')
     spice_frame_pattern = r"(imap)_(?P<version>[\d]+)\.(?P<type>tf)"
 
+    # Covers:
+    # Science Frame: (type: 'tf')
+    science_frame_pattern = r"(imap_science)_(?P<version>[\d]+)\.(?P<type>tf)"
```
tech3371, line 539: You can do something similar to DPS regex.
```suggestion
    science_frame_pattern = r"(?P<type>imap_science)_(?P<version>[\d]+)\.tf"
```

tests/test_file_validation.py
subagonsouth: I'm just trying to figure out how we will distinguish between these two frames kernels when specifying dependencies. For example, I think right now `frames, spice, historical` will get the `imap_NNNN.tf` frames kernel. But how will we specify that we need the science frames kernel?

imap_data_access/file_validation.py
```diff
@@ -534,6 +534,10 @@ class SPICEFilePath(ImapFilePath):
     # Frame: (type: 'tf')
     spice_frame_pattern = r"(imap)_(?P<version>[\d]+)\.(?P<type>tf)"
```
tech3371, line 535: In above `_SPICE_TYPE_MAPPING`, change line 431 to this
```
    "tf": "imap_frames",
    "imap_science": "science_frames",
```

Then update line 453 to this
```
    "imap_frames": "fk",
    "science_frames": "fk",
```

tech3371: @subagonsouth I am looking on this page, https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/228069149/SDC+SPICE+Dependency+Association+SPIKE#SDCSPICEDependencyAssociationSPIKE-SupportedKernelTypes and looks like you had `imap_wkcp.tf` as `imap_frames` is that same as this? if so, you may need to change that to `imap_0000.tf` or something with digits as version.

tests/test_file_validation.py
tech3371: cool, I made above suggestion to differentiate those two from each other. I hope that helps to start.

imap_data_access/file_validation.py
subagonsouth: Yes, that is correct. I believe "wkcp" is for working copy. I have renamed it to `imap_001.tf`, uploaded to s3 and used it in local testing.

tech3371: perfect! Once Sean updates this, he can test querying `imap_frames` or `science_frames` that you wanted in `test_processing_input.py`. He can do something like this
```
    input_collection_str = [
        {"type": "spice", "files": ["imap_001.tf", "imap_science_0001.tf"]},
    ]
    input_collection = processing_input.ProcessingInputCollection()
    input_collection.deserialize(json.dumps(input_collection_str))
    assert len(input_collection.get_file_paths(data_source='science_frames)) == 1
    assert len(input_collection.get_file_paths(data_source='imap_frames)) == 1
```
This will catch anything you missed.

tech3371: yep. Here, we are saying that this input is of type `spice`. was there an error you are seeing with my above suggestion? Let me know on Slack if you like to debug through that.
```
input_collection_str = [
        {"type": "spice", "files": ["imap_001.tf", "imap_science_0001.tf"]},
    ]
```

sdhoyt:
pre-commit.ci autofix

sdhoyt:
pre-commit.ci autofix

imap_data_access/file_validation.py
tech3371: we resolved this on calls. we had to query by data_type and it should return 2 paths since there is two files in 'files' list.

subagonsouth (approved):
Thanks, Sean!

PR #229, opened 2025-05-29
Version bump

maxinelasp:
# Change Summary
Version bump for new release with ancillary querying and CR filename convention.


greglucas:
Looks like the query file needs a `noqa:` added to the function to ignore that check. Although, that is an awful lot of stuff in the query portion now, so it might be useful to refactor that later.

greglucas (approved):

PR #236, opened 2025-06-09
Rename CadenceFilePath to DependencyFilePath

lacoak21:
# Change Summary

## Overview
JSON filepaths containing the serialized processingInputCollection will be passed into every batch job command. Previously, this was only being done for cadence files. This PR is to rename the CadenceFilePath class to the broader DependencyFilePath. 

## Updated Files
All files: 
  - rename cadence -> dependency

### DIFF
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -12,7 +12,7 @@ from pathlib import Path
 
 from imap_data_access.file_validation import (
     AncillaryFilePath,
-    CadenceFilePath,
+    DependencyFilePath,
     ImapFilePath,
     QuicklookFilePath,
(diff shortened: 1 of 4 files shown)


greglucas (approved):
Are there corresponding changes that need to happen in sync here for imap_processing / sds-data-manager? If so, would it be useful to keep a backward compatable shim like:

```python
class CadenceFilePath(DependencyFilePath):
    """DEPRECATED class for CadenceFile processing, use DependencyFilePath instead"""
```

greglucas:
You might still need to add it to the __all__ exports depending on how it is imported from the other packages.

lacoak21:
Good point.

PR #237, opened 2025-06-11
Adding 'End_date' parameter when query by Filename

daralynnrhode:
# Setting End_date Parameter when using filename

## Overview

Fixing a bug that when a user queries by the filename `End_date` never gets sets, so if there is two versions of the file with only differing `start_dates` then both will be returned (if querying by the filename of the earlier date).

## Updated Files

- cli.py


bourque (approved):
Looks good, thanks for fixing this!

PR #238, opened 2025-06-11
Bug: Fixing Index Error for Empty Query Responses

### LINKED ISSUE #232: BUG - Query CLI produces an error when no results returned
### Description of the issue

When no files are found, the cli returns an error "list index out of range" rather than an empty response or something indicating no files were found.

Query for an instrument with no data at that specific level.
```
imap-data-access query --instrument swapi --data-level l3e
IndexError('list index out of range')
```

An empty list or message should be returned with a success response instead of an error.

_No response_

### Code Snippet:

```python
Code
```

_No response_

daralynnrhode:
# Fix for the IndexError Occurring When Query Response is Empty

## Overview

Fixes the bug linked in #232. 
Closes #232 
## Updated Files

- cli.py

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -55,16 +55,17 @@ def _print_query_results_table(query_results: list[dict]):
         A list of dictionaries containing the query results
     """
     num_files = len(query_results)
+    query_table = "science"  # default to science so empty list can be printed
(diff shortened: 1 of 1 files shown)


bourque (approved):
Thank you for fixing this!

PR #243, opened 2025-07-10
ENH: webpoda: Add an instrument-specific buffer for downloading packet files

### LINKED ISSUE #140: Add ability to control buffer while downloading L0 files from webpoda
### Summary

When downloading files from webpoda, we are currently specifying an arbitrary buffer of +/- 30 minutes. We should look into whether we should bring that up into the argument of the function to be controllable, whether it needs to be different on the front/back end. Or if there is some other requirements on instrument processing we need to control.

### Requirements

- [ ] Allow for more control over the length of spacecraft times being downloaded from webpoda

### Related tickets

_No response_

greglucas:
Each instrument can specify the buffer they want to apply to include packets +/- X minutes on either side of the midnight boundary.

closes #140

### DIFF
--- a/imap_data_access/webpoda.py
+++ b/imap_data_access/webpoda.py
@@ -127,6 +127,13 @@ INSTRUMENT_APIDS = {
     "ialirt": [478],
 }
 
+_INSTRUMENT_BUFFER_MINUTES = {
+    "hit": 20,
+    "mag": 30,
+    "swapi": 1,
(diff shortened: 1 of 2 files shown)


tech3371 (approved):
LGTM. Thank you for adding that. And that was quick!

maxinelasp (approved):
Looks great, thank you!

PR #247, opened 2025-07-16
Api key capitalization

greglucas:
The backend wanted "x-api-key" in the header (lowercase x), so change it here since it is unused currently.

We also are adding this to more routes than just upload, so add it generically to any request rather than individually in each routine.

### DIFF
--- a/imap_data_access/io.py
+++ b/imap_data_access/io.py
@@ -32,6 +32,10 @@ def _make_request(request: requests.PreparedRequest):
     when making HTTP requests and yield the response body.
     """
     logger.debug("Making request: %s", request)
+    if imap_data_access.config["API_KEY"]:
+        # Add the API key to the request headers if it exists
+        request.headers["x-api-key"] = imap_data_access.config["API_KEY"]
+
     try:
         with requests.Session() as session:
             response = session.send(request)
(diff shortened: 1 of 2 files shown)


imap_data_access/io.py
```diff
@@ -32,6 +32,10 @@ def _make_request(request: requests.PreparedRequest):
     when making HTTP requests and yield the response body.
     """
     logger.debug("Making request: %s", request)
+    if imap_data_access.config["API_KEY"]:
```
lacoak21, line 35: Nice - just tested with this header key on postman.

imap_data_access/io.py
```diff
@@ -408,8 +412,7 @@ def upload(file_path: Union[Path, str], *, api_key: Optional[str] = None) -> Non
     # We send a GET request with the filename and the server
     # will respond with an s3 presigned URL that we can use
     # to upload the file to the data archive
-    headers = {"X-api-key": api_key} if api_key else {}
-    request = requests.Request("GET", url, headers=headers).prepare()
+    request = requests.Request("GET", url).prepare()
```
lacoak21, line 415: I think you can remove the lines 409-410 now.

lacoak21: And remove the api_key parameter?

lacoak21 (approved):
LGTM just had one comment about the api_key variable in upload()

PR #256, opened 2025-08-12
New Spice Kernel for I-ALiRT

laspsandoval:
# Change Summary

## Overview
This adds the SPICE kernel: 
earth_latest_high_prec.bpc
with naming convention:
earth_000101_yymmdd_yymmdd.bpc

to the list of kernels.

## Updated Files
- file_validation.py
   - Add new kernel

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -477,6 +477,7 @@ _SPICE_TYPE_MAPPING = {
     "tm": "metakernel",
     "sff": "thruster",
     "L1_de": "lagrange_point",
+    "bcp": "high_precision_planetary_constants",
 }
 
 _SPICE_DIR_MAPPING = {
@@ -500,6 +501,7 @@ _SPICE_DIR_MAPPING = {
     "metakernel": "mk",
     "thruster": "activities",
     "lagrange_point": "spk",
+    "high_precision_planetary_constants": "bcp",
 }
 """These are the valid extensions for SPICE files according to NAIF
 https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/kernel.html


imap_data_access/file_validation.py
```diff
@@ -477,6 +477,7 @@ def is_valid_cr(input_cr: str) -> bool:
     "tm": "metakernel",
     "sff": "thruster",
     "L1_de": "lagrange_point",
+    "bcp": "high_precision_planetary_constants",
```
subagonsouth, line 480: This is for `earth_latest_high_prec.bpc` correct? .bpc files are generically called planetary CK files or to fit with our other file names, perhaps we should use planetary_attitude, or more specifically earth_attitude.
```suggestion
    "bpc": "earth_attitude",
```

imap_data_access/file_validation.py
```diff
@@ -500,6 +501,7 @@ def is_valid_cr(input_cr: str) -> bool:
     "metakernel": "mk",
     "thruster": "activities",
     "lagrange_point": "spk",
+    "high_precision_planetary_constants": "bcp",
```
subagonsouth, line 504: It looks like the standard directory for planetary constants files is "pck".
```suggestion
    "earth_attitude": "pck",
```

subagonsouth (commented):
It looks like there is some confusion in what this kernel provides. See my suggested changes.

imap_data_access/file_validation.py
```diff
@@ -500,6 +501,7 @@ def is_valid_cr(input_cr: str) -> bool:
...
 }
 """These are the valid extensions for SPICE files according to NAIF
 https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/kernel.html
```
tech3371, line 507: Also, after Tim's suggestion, you need to add regex for that filename convention and add a test.

imap_data_access/file_validation.py
```diff
@@ -621,6 +624,13 @@ class SPICEFilePath(ImapFilePath):
         r"e(?P<version>\d{2})\."
         r"(?P<type>mk)"
     )
+    earth_attitude_pattern = (
```
subagonsouth, line 627: Did you decide to use the kernels named with start, stop, predict dates so that coverage dates could be parsed from the name? I think that we requested the MOC to get only the `earth_latest_high_prec.bpc` kernel.

imap_data_access/file_validation.py
```diff
@@ -621,6 +624,13 @@ class SPICEFilePath(ImapFilePath):
...
+        r"earth_"
+        r"(?P<start_date>\d{6})_"
+        r"(?P<end_date>\d{8})_"
+        r"(?P<prediction_start_date>\d{8})"
```
subagonsouth, line 631: NAIF says that the filename pattern is: `earth_000101_yymmdd_yymmdd.bpc` which would mean that you want:
```suggestion
        r"(?P<end_date>\d{6})_"
        r"(?P<prediction_start_date>\d{6})"
```

tests/test_file_validation.py
```diff
@@ -391,6 +391,17 @@ def test_spice_extract_science_frame_parts():
     assert len(file_path.spice_metadata) == 5
 
 
+def test_spice_extract_earth_attitude_parts():
+    file_path = SPICEFilePath("earth_000101_20251120_20260220.bpc")
```
subagonsouth, line 395: As I pointed out above, this filename does not match the NAIF convention. Can you use actual filenames found here for testing?
https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/

subagonsouth (commented):
I'm curious to hear your thoughts on the change to not use the `earth_latest_high_prec.bcp` kernel. I am not opposed to the change, just want to make sure that the added complexity is worth it.

PR #266, opened 2025-09-05
Added support for .dat extension science input files

pleasant-menlo:
# Change Summary

## Overview
Added support for .dat extension science input files. This is for support for GLOWS L3D which outputs both CDF files and DAT files. 

## Updated Files

- file_validation.py
   - added "dat" to the set of valid extensions
- test_file_validation.py
   - updated tests to reflect dat support in production
- test_io.py
   - error now includes dat as a "choose from" in the set of valid extensions

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -127,7 +127,7 @@ class ScienceFilePath(ImapFilePath):
         "<mission>_<instrument>_<datalevel>_<descriptor>_"
         "<startdate>(-<repointing>)_<version>.<extension>"
     )
-    VALID_EXTENSIONS: typing.ClassVar[set[str]] = {"cdf", "pkts"}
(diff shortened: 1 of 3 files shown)


pleasant-menlo:
@greglucas @maxinelasp @lacoak21 Is this all that is needed to allow for .dat science files to be uploaded/downloaded?

maxinelasp (changes requested):
I agree with Greg, if my understanding of L3 is correct, you actually want to output these .dat files as ancillary files. "Science files" refer specifically to the archived CDF or PKTS files which external users will need only. Science team only or internal files, even if output by processing, have always been considered ancillary files.

maxinelasp:
@pleasant-menlo Unfortunately, if it's anything like the science files, that change would require touching a lot of different components since it requires a database change. However, you can include the cr in the descriptor of the files - `imap_glows_e-dens-cr02094_19470303_v001.dat`

PR #278, opened 2025-10-07
FIX: Handle potential NaNs in the middle of the repointing files

greglucas:
We are getting files with NaN values in the middle and that is causing the first conversion to date of the start_time to fail now. Check for that case and skip over any repointings with an invalid start date.


subagonsouth (approved):
Looks good.

PR #280, opened 2025-10-08
Support for access token authorization

pleasant-menlo:
Added support for using an access token for authorization when making requests.

-Giovanni and George


greglucas:
These commits and the other documentation fix I pushed were merged in #280, so this should be taken care of now.

PR #286, opened 2025-11-05
Minor API Key Readme Update

sdhoyt:
# Change Summary

## Overview
In the documentation for setting the API Key, I moved the environment variable setting step to a line before using the `imap-data-access`, so it's more clear that it's not part of the `imap-data-access` command. I also added `export` for setting the environment variable

PR #288, opened 2025-12-04
Clarifying readme for auth keys

maxinelasp:
# Change Summary
Updated readme to clarify some confusion I had

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,14 +144,13 @@ There are two ways to provide your API Key:
 1. **Environment variable**: Set the `IMAP_API_KEY` environment variable
 
 ```bash
-$ export IMAP_API_KEY=<your-api-key>
-$ imap-data-access ...
+IMAP_API_KEY=<your-api-key> imap-data-access ...
 ```
 
 2. **Command line flag**: Use the `--api-key` flag
 
 ```bash
-$ imap-data-access --api-key <your-api-key> ...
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,28 +132,34 @@ the environment variable ``IMAP_DATA_ACCESS_URL`` or within the
...
 
-To use the API Key you can set the `IMAP_API_KEY` environment variable and then use
-the tool as usual.
+There are two ways to provide your API Key:
+
+1. **Environment variable**: Set the `IMAP_API_KEY` environment variable
 
 ```bash
-$ IMAP_API_KEY=<your-api-key>
+$ export IMAP_API_KEY=<your-api-key>
```
greglucas: I still have a slight preference for doing this on the same line like we had it before instead of an export (which is also shell specific).
https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/286#issuecomment-3498566753

But, I agree this does need to change.

greglucas (approved):
I also missed the `$` additions in the previous PR. I don't love those because then I can't copy-paste. Because it is in a code block I don't think we need the `$ my-command` prefix. But there is apparently disagreement on preferences here too :) https://github.com/orgs/community/discussions/35615

All in all, this is an improvement and looks good to me! Feel free to take or leave the suggestions at your own will, I'm fine with any of it.

PR #308, opened 2026-05-04
Feature: Adding data release filename convention

### LINKED ISSUE #305: Release: Add filename convention of release files
Based on notes on this page, add object for release files and their s3 path in file_validation.py, similar to ancillary and other file types.

https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/324168272/Public+Releases

tech3371:
# Change Summary
closes #305

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -34,6 +34,7 @@ def generate_imap_file_path(filename: str) -> ImapFilePath:
         SPICEFilePath,
         QuicklookFilePath,
         DependencyFilePath,
+        ReleaseFilePath,
     ):
         try:
             return cls(filename)
@@ -1013,8 +1014,8 @@ class AncillaryFilePath(ImapFilePath):
         upload_path = Path(f"{self._dir_prefix}/{self.instrument}/{self.filename}")
         return imap_data_access.config["DATA_DIR"] / upload_path
 
-    @staticmethod
-    def extract_filename_components(filename: str | Path) -> dict:
+    @classmethod
+    def extract_filename_components(cls, filename: str | Path) -> dict:
         """Extract all components from filename. Does not validate instrument or level.
 
         Will return a dictionary with the following keys:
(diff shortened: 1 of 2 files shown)


imap_data_access/file_validation.py
```diff
@@ -1109,3 +1110,10 @@ class DependencyFilePath(ScienceFilePath):
 
 class CadenceFilePath(DependencyFilePath):
     """DEPRECATED class for CadenceFile processing, use DependencyFilePath instead."""
+
+
+class ReleaseFilePath(AncillaryFilePath):
+    """Release filename convention."""
+
+    VALID_EXTENSIONS: typing.ClassVar[set[str]] = {"txt"}
+    _dir_prefix = "imap/release"
```
vmartinez-cu, line 1119: I was originally thinking about inheriting the AncillaryFilePath class as well but I wonder if issues could arise in the future since any future ancillary parsing changes would ripple into releases. Any concerns about this?

tech3371: I don't see ancillary filename convention changing. If so, that will be a big one :)

tests/test_file_validation.py
```diff
@@ -697,3 +698,34 @@ def test_dependency_file_path():
...
+        "imap_mag_withhold-data-release-001_20260201_20260430_v001.txt"
+    )
+    assert release_file_mag.instrument == "mag"
+    assert release_file_mag.descriptor == "withhold-data-release-001"
+    assert release_file_mag.start_date == "20260201"
+    assert release_file_mag.end_date == "20260430"
+
+    # Test selected-release descriptor with date range
+    release_file_swe = ReleaseFilePath(
+        "imap_swe_selected-release_20260201_20260430_v001.txt"
```
vmartinez-cu: How is `selected-release` different from `withhold-data`? Does "withhold" cover all files for a specific release date? Here's what I found in our documentation but it's not super clear what scenarios would require one over the other

```
withhold-data-release-<###> - it will support integer value associated to a release number from above table.

selected-release  - it will support making release of selected files
```

tech3371: withhold-data are ones that happens during mission wide release. But I am visioning getting requests to release some files sooner due to paper being published or something. That allows selected releases which will be subset.

tech3371: one of the reason I didn't narrow down release filename convention's descriptor is because I wanted to leave it pretty open and flexible for now to allow us to figure out the details in `sds` repo before locking it down.

vmartinez-cu (approved):

PR #316, opened 2026-05-18
ENH: Adding Release command to IMAP CLI

### LINKED ISSUE #315: ENH: add CLI command for data release API
### Summary

Similar to `upload` and `query`, add `release` command and support

### Requirements

The more specific the better.
  - [ ] Support `imap-data-access release --instrument --start-date --end-date --release-type --release-number (optional)`
  - [ ]

### Related tickets

_No response_

tech3371:
# Change Summary
closes #315 

## Overview

Ability to make release through IMAP CLI.

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -357,7 +357,7 @@ def _release_parser(args: argparse.Namespace):
         release_number=getattr(args, "release_number", None),
         table=getattr(args, "table", None),
         descriptor=getattr(args, "descriptor", None),
-        manifest_file_path=getattr(args, "manifest_file", None),
+        manifest_file=getattr(args, "manifest_file", None),
     )
     print("Successfully submitted release request to the IMAP SDC.")
 
@@ -740,10 +740,9 @@ def main():
         required=False,
         metavar="PATH",
         help=(
-            "Path to manifest file containing list of files. 'release' type"
-            " is mission wide release and doesn't require a manifest file. "
-            "Required for 'early-release' and 'unrelease' release types. "
-            "Be sure manifest file follows release filename convention."
+            "Path to manifest file containing list of files to "
+            "withhold release during mission wide release or "
+            "early-release or unrelease."
         ),
     )
     parser_release.set_defaults(func=_release_parser)
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -357,7 +357,7 @@ def _release_parser(args: argparse.Namespace):
         release_number=getattr(args, "release_number", None),
         table=getattr(args, "table", None),
         descriptor=getattr(args, "descriptor", None),
-        manifest_file=getattr(args, "manifest_file", None),
+        exception_list_file=getattr(args, "exception_list", None),
     )
     print("Successfully submitted release request to the IMAP SDC.")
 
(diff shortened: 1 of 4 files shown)


imap_data_access/cli.py
```diff
@@ -662,6 +685,68 @@ def main():
...
+    parser_release.add_argument(
+        "--release-number",
+        type=int,
+        required=False,
+        metavar="NUMBER",
+        help="Release number (required only when --release-type is 'release'). ",
+    )
+
+    parser_release.add_argument(
+        "--manifest-file",
```
tmplummer: I think that this parameter could be named more clearly. Without reading the help string, I would think that a manifest file would replace the start/end date and be a list of files that the user wants the action (release, unrelease, early-release) to act on. Naming this something like `--exception-list` would be more intuitive.

imap_data_access/cli.py
```diff
@@ -662,6 +685,68 @@ def main():
...
+        required=True,
+        metavar="YYYYMMDD",
+        help="End date for the release",
+    )
+    parser_release.add_argument(
+        "--release-type",
+        type=str,
+        required=True,
+        metavar="ReleaseType",
+        help="Type of release: 'release', 'unrelease', 'early-release'",
```
tmplummer: More detail specifying what each of the three choices do would be nice.

imap_data_access/cli.py
```diff
@@ -662,6 +685,68 @@ def main():
...
+
+    parser_release.add_argument(
+        "--manifest-file",
+        type=str,
+        required=False,
+        metavar="PATH",
+        help=(
+            "Path to manifest file containing list of files to "
+            "withhold release during mission wide release or "
+            "early-release or unrelease."
```
tmplummer: I think that this could be more clearly stated:
"Path to a manifest file containing a list of files to exclude from the action specified by the --release-type parameter."

tmplummer (commented):
Just a couple of things to consider.

imap_data_access/cli.py
tech3371: I like that name!

imap_data_access/cli.py
```diff
@@ -662,6 +685,83 @@ def main():
...
+        type=str,
+        required=False,
+        metavar="PATH",
+        help=(
+            "Path to exception list file containing list of files to exclude\n"
+            "or include based on release type:\n"
+            "- For 'release': files to withhold from public release\n"
+            "- For 'early-release': files to release early\n"
+            "- For 'unrelease': files to unrelease from previously released\n"
+            "  data"
```
tmplummer: Oh, interesting. I am glad I commented. This is not what I expected. Due to the different handling of this list of files from one release type to the next, would it be worth breaking it into different arguments. For release, the list of files to release is: "all files in the specified date range minus the files listed. But, for early-release and unrelease, the list of files here is the list to act on. Maybe this is too nit-picky, but I think it is good to have consistency in the UI to avoid confusion.

To add to this, it seems like for early-release and unrelease the date arguments are not required?

tmplummer (approved):
I know that we are in a crunch, so I won't hold this up, but I think the parameters for the release CLI are still a bit confusing and inconsistent.

imap_data_access/cli.py
tech3371: I am planning to not support early-release and unrelease options yet on the lambda due to time crunch. Should I remove them for now?

tmplummer (approved):
Looks good!

PR #321, opened 2026-06-03
ENH: adding config with data levels and descript for imap mission

tech3371:
# Change Summary

## Overview

We have been getting more frequent updates to website drop-down menu selections. It's been hard to track changes and pass information to web team.

We have been meaning to add valid data levels and descriptor for each instrument once it's finalized. We couldn't before but now may be good time to add these finalized information. I am hopeful that it will change less frequent in the future.

I updated on this repo and library since web team may be using this already. Repo `sds-data-manager` is too big and irrelevant to web team and other users.

With this update, website team can update `imap-data-access` version and call these new class methods to get valid data levels, descriptors for each data levels and ancillary descriptor and etc.

## File changes

- added json config file with tree of science and ancillary descriptors.
- add function to read in those information based on use cases for website and may be more users.

### DIFF
--- a/imap_data_access/__init__.py
+++ b/imap_data_access/__init__.py
@@ -28,6 +28,7 @@ from imap_data_access.processing_input import (
     SPICEInput,
     SpinInput,
 )
+from imap_data_access.utils import ImapProductCatalog
 
 __all__ = [
     "VALID_DATALEVELS",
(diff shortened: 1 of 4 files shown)


tech3371:
After chatting with Ransom from web team, this won't help them. web team only uses API. Closing this PR for now

PR #319, opened 2026-05-28
Update Query version latest filtering

pleasant-menlo:
## Change Summary
Before this change, querying with version=latest only returned one file per start date. The desired behavior is that this should return multiple files for each day if the query returns files with multiple descriptors, or if there are multiple repointings in a single day.

Querying with version latest now only excludes a file if there is a higher-version file with matching instrument, data level, descriptor, start_date, and repointing.

```
imap-data-access query --instrument glows --data-level l3e --descriptor survival-probability-hi-45 --version latest --start-date 20260407 --end-date 20260408
```

Used to return

```
 imap_glows_l3e_survival-probability-hi-45_20260407-repoint00209_v006.cdf
 imap_glows_l3e_survival-probability-hi-45_20260408-repoint00212_v006.cdf 
```
And now returns
```
imap_glows_l3e_survival-probability-hi-45_20260407-repoint00209_v006.cdf
imap_glows_l3e_survival-probability-hi-45_20260407-repoint00210_v001.cdf
imap_glows_l3e_survival-probability-hi-45_20260407-repoint00211_v001.cdf
imap_glows_l3e_survival-probability-hi-45_20260408-repoint00212_v006.cdf

```

We believe this closes #289

PR #324, opened 2026-06-08
ScienceFilePath release version format

### LINKED ISSUE #320: ENH: update science filename convention to support `vRRR.MMM`
### Summary

We need to update `ScienceFilePath` to support this new version schema. 

### Requirements

The more specific the better.
  - [x] Update `ScienceFilePath` to support `vRRR.MMM`
  - [x] Update `ScienceFilePath` to say that `vMMM` is `DEPRECATED` but support it for backward compatible while we make update
  - [x] Double check that `ScienceInput(ProcessingInput)` won't be effected due to above updates and update as needed.

### Related tickets

_No response_

lacoak21:
# Change Summary

closes #320 

## Overview
Update the ScienceFilePath version to contain release number information. Add backwards compatibility for old version format while we transition. 

## File changes
  - Update version

## Testing
Fix tests and add new ones for new version format.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -127,6 +127,12 @@ class ScienceFilePath(ImapFilePath):
     FILENAME_CONVENTION = (
         "<mission>_<instrument>_<datalevel>_<descriptor>_"
         "<startdate>(-<repointing>)_<version>.<extension>"
+        " where version is vRRR.MMM (legacy vXXX is deprecated but supported)"
+    )
+    NEW_VERSION_PATTERN: typing.ClassVar[str] = r"v\d{3}\.\d{3}"
+    DEPRECATED_VERSION_PATTERN: typing.ClassVar[str] = r"v\d{3}"
+    VALID_VERSION_PATTERN: typing.ClassVar[str] = (
+        rf"(?:{NEW_VERSION_PATTERN}|{DEPRECATED_VERSION_PATTERN})"
     )
     VALID_EXTENSIONS: typing.ClassVar[set[str]] = {"cdf", "pkts"}
     _dir_prefix = "imap"
@@ -158,7 +164,8 @@ class ScienceFilePath(ImapFilePath):
             repointing the data is from, format: repointXXXXX
         <cr>: This is an optional field describing the Carrington rotation.
             format: crXXXXX.
-        <version>: This stores the data version for this product, format: vXXX
+        <version>: This stores the data version for this product, format: vRRR.MMM.
+            Legacy vXXX is accepted for backward compatibility but deprecated.
 
         Parameters
         ----------
@@ -290,14 +297,6 @@ class ScienceFilePath(ImapFilePath):
                 f"Invalid filename, missing attribute. Filename "
                 f"convention is {ScienceFilePath.FILENAME_CONVENTION} \n"
             )
-        # TODO remove this check once version is updated from vXXX -> vRRR.MMM and add
-        #   self.release_number and self.data_version to the list above.
-        if "." in self.version and (
-            self.release_number is None or self.data_version is None
-        ):
-            error_message = (
-                "Invalid version, missing attribute. Version convention is vRRR.MMM"
-            )
         if self.mission != "imap":
             error_message += f"Invalid mission {self.mission}. Please use imap \n"
 
@@ -315,12 +314,10 @@ class ScienceFilePath(ImapFilePath):
             )
         if not self.is_valid_date(self.start_date):
             error_message += "Invalid start date format. Please use YYYYMMDD format. \n"
-        # TODO update pattern to be r"^(v\d{3}\.\d{3})$" once versions are
-        #  updated from vXXX to vRRR.MMM
-        if not bool(re.match(r"^(v\d{3}|v\d{3}\.\d{3})$", self.version)):
+        if not ScienceFilePath.is_valid_version(self.version):
             error_message += (
                 "Invalid version format. Please use vRRR.MMM format"
-                " (vXXX format is deprecated).\n"
+                " (vXXX format is deprecated but supported for compatibility).\n"
             )
         if self.repointing and not isinstance(self.repointing, int):
             error_message += "The repointing number should be an integer.\n"
@@ -384,9 +381,7 @@ class ScienceFilePath(ImapFilePath):
             r"(?P<start_date>\d{8})"
             # Optional repointing/CR field
             r"(-(?P<interval_type>(?:repoint|cr))(?P<interval>\d{5}))?"
-            # TODO update pattern to be r"_(?P<version>v\d{3}\.\d{3})" once versions are
-            #  updated from vXXX to vRRR.MMM
-            r"_(?P<version>v\d{3}|v\d{3}\.\d{3})"
+            rf"_(?P<version>{ScienceFilePath.VALID_VERSION_PATTERN})"
             r"\.(?P<extension>[^.]+)$"
         )
         if isinstance(filename, Path):
@@ -466,20 +461,13 @@ class ScienceFilePath(ImapFilePath):
 
     @staticmethod
     def is_valid_version(input_version: str) -> bool:
-        """Check input version string is in valid format 'vXXX' or 'latest'.
+        """Check input version string is valid for science files.
 
-        Parameters
-        ----------
-        input_version : str
-            Version to be checked.
-
-        Returns
-        -------
-        bool
-            Whether input version is valid or not.
+        A valid science version is ``vRRR.MMM`` or legacy ``vXXX``. The special value
+        ``latest`` is also accepted for query-style inputs.
         """
-        return input_version == "latest" or re.fullmatch(
-            r"(v\d{3}|v\d{3}\.\d{3})", input_version
+        return input_version == "latest" or bool(
+            re.fullmatch(ScienceFilePath.VALID_VERSION_PATTERN, input_version)
         )
 
 
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -460,6 +460,22 @@ class ScienceFilePath(ImapFilePath):
             return False
 
     @staticmethod
+    def is_valid_cr(input_cr: str) -> bool:
+        """Check input carrington rotation string is in valid format 'crXXXXX'.
+
+        Parameters
+        ----------
+        input_cr : str
+            Carrington rotation to be checked.
+
+        Returns
+        -------
+        bool
+            Whether input carrington rotation is valid or not.
+        """
+        return re.fullmatch(r"cr\d{5}", str(input_cr))
+
+    @staticmethod
     def is_valid_version(input_version: str) -> bool:
         """Check input version string is valid for science files.
 
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -127,9 +127,9 @@ class ScienceFilePath(ImapFilePath):
     FILENAME_CONVENTION = (
         "<mission>_<instrument>_<datalevel>_<descriptor>_"
         "<startdate>(-<repointing>)_<version>.<extension>"
-        " where version is vRRR.MMM (legacy vXXX is deprecated but supported)"
+        " where version is vMMM.mmmm (legacy vXXX is deprecated but supported)"
     )
-    NEW_VERSION_PATTERN: typing.ClassVar[str] = r"v\d{3}\.\d{3}"
+    NEW_VERSION_PATTERN: typing.ClassVar[str] = r"v\d{3}\.\d{4}"
     DEPRECATED_VERSION_PATTERN: typing.ClassVar[str] = r"v\d{3}"
     VALID_VERSION_PATTERN: typing.ClassVar[str] = (
         rf"(?:{NEW_VERSION_PATTERN}|{DEPRECATED_VERSION_PATTERN})"
@@ -164,7 +164,7 @@ class ScienceFilePath(ImapFilePath):
             repointing the data is from, format: repointXXXXX
         <cr>: This is an optional field describing the Carrington rotation.
             format: crXXXXX.
-        <version>: This stores the data version for this product, format: vRRR.MMM.
+        <version>: This stores the data version for this product, format: vMMM.mmmm.
             Legacy vXXX is accepted for backward compatibility but deprecated.
 
         Parameters
@@ -190,8 +190,8 @@ class ScienceFilePath(ImapFilePath):
         self.repointing = split_filename["repointing"]
         self.cr = split_filename["cr"]
         self.version = split_filename["version"]
-        self.release_number = split_filename["release_number"]
-        self.data_version = split_filename["data_version"]
+        self.major_version = split_filename["major_version"]
+        self.minor_version = split_filename["minor_version"]
         self.extension = split_filename["extension"]
 
         self.error_message = self.validate_filename()
@@ -316,7 +316,7 @@ class ScienceFilePath(ImapFilePath):
             error_message += "Invalid start date format. Please use YYYYMMDD format. \n"
         if not ScienceFilePath.is_valid_version(self.version):
             error_message += (
-                "Invalid version format. Please use vRRR.MMM format"
+                "Invalid version format. Please use vMMM.mmmm format"
                 " (vXXX format is deprecated but supported for compatibility).\n"
             )
         if self.repointing and not isinstance(self.repointing, int):
@@ -412,16 +412,16 @@ class ScienceFilePath(ImapFilePath):
         del components["interval_type"]
 
         # Initialize version components
-        components["release_number"] = None
-        components["data_version"] = None
+        components["major_version"] = None
+        components["minor_version"] = None
         version = components["version"]
         # TODO update this logic to be more strict once versions are updated from
-        #  vXXX to vRRR.MMM. Right now, we are just checking if there's a "." in the
+        #  vXXX to vMMM.mmmm. Right now, we are just checking if there's a "." in the
         #  version to determine if it's in the new format or old format.
         if "." in version:
-            release_number, data_version = version.split(".")
-            components["release_number"] = int(release_number[1:])  # Remove the "v"
-            components["data_version"] = int(data_version)
+            major_version, minor_version = version.split(".")
+            components["major_version"] = int(major_version[1:])  # Remove the "v"
+            components["minor_version"] = int(minor_version)
 
         return components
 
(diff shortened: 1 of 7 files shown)


imap_data_access/file_validation.py
```diff
@@ -436,20 +460,15 @@ def is_valid_for_start_date(self, start_date: datetime) -> bool:
             return False
 
     @staticmethod
-    def is_valid_cr(input_cr: str) -> bool:
```
tech3371, line 439: we shouldn't remove this `is_valid_cr()` function.

imap_data_access/io.py
```diff
@@ -390,16 +423,16 @@ def query(
...
-                version_num > latest_per_day[day]["_version_num"]
+                version_key > latest_per_day[day]["version_key"]
             ):
                 # add extra field to identify version number
-                latest_per_day[day] = {**item, "_version_num": version_num}
+                latest_per_day[day] = {**item, "version_key": version_key}
         # remove the extra field
         items = [
-            {k: version for k, version in val.items() if k != "_version_num"}
+            {k: version for k, version in val.items() if k != "version_key"}
```
tech3371: can you review this PR since they are closely related? https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/319

If that PR logic looks good, can you merge that PR and rebase with this?

tech3371 (commented):
I had couple comments but overall looks good to me!

imap_data_access/file_validation.py
lacoak21: Oh oops! I didnt mean to remove that.

imap_data_access/webpoda.py
```diff
@@ -536,11 +536,25 @@ def _get_latest_version_file_path(
         end_date=start_time.strftime("%Y%m%d"),
         repointing=repointing,
     )
+    # TODO query for release number and replace with variable below.
+    release_number = 1
```
tech3371: can you use default as 0 to be safe?

imap_data_access/io.py
```diff
@@ -398,12 +431,13 @@ def get_key(file_entry):
...
+        version_sort_keys = {}
         for item in items:
             key = get_key(item)
-            if (key not in latest_files) or (
-                item["version"] > latest_files[key]["version"]
-            ):
+            version_key = _version_sort_key(item["version"])
+            if (key not in latest_files) or (version_key > version_sort_keys[key]):
                 latest_files[key] = item
+                version_sort_keys[key] = version_key
```
tech3371: should this be `latest_version`? Similar to `latest_files`
```suggestion
                latest_versions[key] = version_key
```

tech3371 (approved):
looks good to me. Thank you for looking at all places and finding holes to fill! I forgot about webpoda and nice find!

imap_data_access/webpoda.py
lacoak21: This is temporary and wont be merged in. I need to make the api release table query then I will update this to be from the table.

imap_data_access/io.py
```diff
@@ -221,6 +228,32 @@ def _validate_query_parameters(**kwargs) -> None:
             )
 
 
+def _version_sort_key(version) -> tuple[int, int]:
```
maxineofficial: This seems like it might be useful outside of these functions, does it make sense to make it a public method? Or even a property or something like that?

imap_data_access/file_validation.py
```diff
@@ -127,6 +127,12 @@ class ScienceFilePath(ImapFilePath):
     FILENAME_CONVENTION = (
         "<mission>_<instrument>_<datalevel>_<descriptor>_"
         "<startdate>(-<repointing>)_<version>.<extension>"
+        " where version is vRRR.MMM (legacy vXXX is deprecated but supported)"
+    )
+    NEW_VERSION_PATTERN: typing.ClassVar[str] = r"v\d{3}\.\d{3}"
```
tech3371:
```suggestion
        " where version is MMM.mmmm (legacy vXXX is deprecated but supported)"
    )
    NEW_VERSION_PATTERN: typing.ClassVar[str] = r"v\d{3}\.\d{4}"
```

imap_data_access/file_validation.py
```diff
@@ -183,6 +190,8 @@ def __init__(self, filename: str | Path):
         self.repointing = split_filename["repointing"]
         self.cr = split_filename["cr"]
         self.version = split_filename["version"]
+        self.release_number = split_filename["release_number"]
+        self.data_version = split_filename["data_version"]
```
tech3371:
```suggestion
        self.major_version = split_filename["major_version"]
        self.minor_version = split_filename["minor_version"]
```

imap_data_access/webpoda.py
```diff
@@ -536,11 +536,24 @@ def _get_latest_version_file_path(
         end_date=start_time.strftime("%Y%m%d"),
         repointing=repointing,
     )
+    # TODO query for release number and replace with variable below.
+    release_number = imap_data_access.query_release_versions()["release_number"]
+    release_str = f"v{release_number:03}"
```
tech3371: Now, we don't need to query for release number or major version. We need to rename `release_number` to `major_version` and default L0 file's major_version to 1 per SDC team discussion.

tech3371 (approved):
I made minor suggestions for new major and minor outcomes. This PR still applies nicely.

vineetbansal:
@lacoak21 - I'm enumerating all the places in `imap_processing` where a specific format of the version string is assumed, so that going forward these call sites can instead call some convenient functions/methods in `imap-data-access`. This is just to inform us on what additional functionality might be useful to expose as part of this PR.


1. `cli.py`, lines 570-572 in `post_processing`:
```
        r = re.compile(r"v\d{3}")
        if not isinstance(self.version, str) or r.match(self.version) is None:
            self.version = f"v{int(self.version):03d}"  # vXXX
```

2. `cli.py` line 581 in `post_processing`:
```
        ds.attrs["Data_version"] = self.version[1:]  # Strip 'v' from version
```

3. `cdf/utils.py`, lines 121-132, in `write_cdf`:
```
    version = dataset.attrs.get("Data_version", None)
    if version is None:
        warnings.warn(
            "No Data_version attribute found in dataset. Using default v999.",
            stacklevel=2,
        )
        version = "999"
        dataset.attrs["Data_version"] = version
    elif not re.match(r"\d{3}", version):
        raise ValueError(
            f"The Data_version attribute {version} does not match expected format XXX."
        )
```

4. `cdf/utils.py`, lines 142, in `write_cdf`:
```
    science_file = imap_data_access.ScienceFilePath.generate_from_inputs(
        instrument=instrument,
        data_level=data_level,
        descriptor=descriptor,
        start_time=start_date,
        version=f"v{version}",  # Ensure version is prefixed with 'v'
        repointing=repointing_int,
    )
```

5. `cdf/utils.py`, lines 221, in `parse_filename_like`:
```
        r"(?:_(?P<version>v\d{3}))?"  # Optional version
```

6. `ancillary/ancillary_dataset_combiner.py`, line 234:
```
        sorted_data_list = sorted(
            self.timestamped_data, key=lambda x: int(x.version[-3:])
        )
```

7. `ancillary/ancillary_dataset_combiner.py`, line 281:
```
                        output_dataset["input_file_version"].data[index] = int(
                            data_input.version[-3:]
                        )
```

Given these (assuming I haven't missed any), I think it would be nice for this PR to provide functionality for:

1. parsing - some way of taking an int/prefixed_string/unprefixed_string, and return a canonical version string (with leading zeros).
2. removing prefixes - something that can be used in CDFs, so functionality that removes the `v` prefix given a version.
3. get separate major and minor versions as ints.
4. ordering - so that callers are not doing their own ordering. (this may not be needed since only ancillary files are doing this ordering, which I understand will not be renamed).
5. validation - `is_valid()` or something similar.
6. A named regex group that callers can use (like in 5 above).
7. Some concept of "unknown" or "missing" version - so `imap_processing` is not doing something like `version = "999"` above.

Perhaps introducing a `Version` class that provides all of these things might be a clean way to do it, which this PR can then also use internally.

Of course this functionality is not a must-have, but then we'll be duplicating a lot of this versioning logic in `imap_processing`.

lacoak21:
> I made minor suggestions for new major and minor outcomes. This PR still applies nicely.

Thanks! Yep I outlined what needed to be changed during the meeting.

lacoak21:
> @lacoak21 - I'm enumerating all the places in `imap_processing` where a specific format of the version string is assumed, so that going forward these call sites can instead call some convenient functions/methods in `imap-data-access`. This is just to inform us on what additional functionality might be useful to expose as part of this PR.
> 
> 1. `cli.py`, lines 570-572 in `post_processing`:
> 
> ```
>         r = re.compile(r"v\d{3}")
>         if not isinstance(self.version, str) or r.match(self.version) is None:
>             self.version = f"v{int(self.version):03d}"  # vXXX
> ```
> 
> 2. `cli.py` line 581 in `post_processing`:
> 
> ```
>         ds.attrs["Data_version"] = self.version[1:]  # Strip 'v' from version
> ```
> 
> 3. `cdf/utils.py`, lines 121-132, in `write_cdf`:
> 
> ```
>     version = dataset.attrs.get("Data_version", None)
>     if version is None:
>         warnings.warn(
>             "No Data_version attribute found in dataset. Using default v999.",
>             stacklevel=2,
>         )
>         version = "999"
>         dataset.attrs["Data_version"] = version
>     elif not re.match(r"\d{3}", version):
>         raise ValueError(
>             f"The Data_version attribute {version} does not match expected format XXX."
>         )
> ```
> 
> 4. `cdf/utils.py`, lines 142, in `write_cdf`:
> 
> ```
>     science_file = imap_data_access.ScienceFilePath.generate_from_inputs(
>         instrument=instrument,
>         data_level=data_level,
>         descriptor=descriptor,
>         start_time=start_date,
>         version=f"v{version}",  # Ensure version is prefixed with 'v'
>         repointing=repointing_int,
>     )
> ```
> 
> 5. `cdf/utils.py`, lines 221, in `parse_filename_like`:
> 
> ```
>         r"(?:_(?P<version>v\d{3}))?"  # Optional version
> ```
> 
> 6. `ancillary/ancillary_dataset_combiner.py`, line 234:
> 
> ```
>         sorted_data_list = sorted(
>             self.timestamped_data, key=lambda x: int(x.version[-3:])
>         )
> ```
> 
> 7. `ancillary/ancillary_dataset_combiner.py`, line 281:
> 
> ```
>                         output_dataset["input_file_version"].data[index] = int(
>                             data_input.version[-3:]
>                         )
> ```
> 
> Given these (assuming I haven't missed any), I think it would be nice for this PR to provide functionality for:
> 
> 1. parsing - some way of taking an int/prefixed_string/unprefixed_string, and return a canonical version string (with leading zeros).
> 2. removing prefixes - something that can be used in CDFs, so functionality that removes the `v` prefix given a version.
> 3. get separate major and minor versions as ints.
> 4. ordering - so that callers are not doing their own ordering. (this may not be needed since only ancillary files are doing this ordering, which I understand will not be renamed).
> 5. validation - `is_valid()` or something similar.
> 6. A named regex group that callers can use (like in 5 above).
> 7. Some concept of "unknown" or "missing" version - so `imap_processing` is not doing something like `version = "999"` above.
> 
> Perhaps introducing a `Version` class that provides all of these things might be a clean way to do it, which this PR can then also use internally.
> 
> Of course this functionality is not a must-have, but then we'll be duplicating a lot of this versioning logic in `imap_processing`.

Thanks Vineet for those thoughts! That is a great idea. I will go through these and see where they fit within this PR.

lacoak21:
@vineetbansal looks like a lot of those could use ScienceFilePath.is_valid_version() which is already implemented in this PR. It uses the VALID_VERSION_PATTERN so seems like 5 and 6 are taken care of unless i am mistaken. I will try to capture the rest in more functions. What gets tricky is that this new version format is only used for science files and NOT ancillary or spice files so we have to be aware of that in imap_processing.

vineetbansal:
@lacoak21 - yes - `is_valid_version(version: str, raise: bool = True) -> bool` should capture a lot of these. Other than that:

- `to_version(major: str|int|None, minor: str|int, raise: bool = True) -> str | None`
- `from_version(version: str, raise: bool = True) -> tuple[str|None, str] | None`
- `version_regex() -> str`

should cover most use cases.

`raise` argument suggested here so callers are not creating their own exception messages, if `False`, then functions can return `None`.

The whole concept of major/minor can be hidden from callers by using a new `Version` class and only letting its constructor be aware of these, and let the functions work exclusively with this class, but that might complicate this PR too much at this point, so probably an overkill.

EDIT: The functions should probably just return well-formatted-and-padded `str`s, while what they take in (other than the `is_valid_version` case) can be a bit more relaxed, like `int`s or strings without zero-padding.

lacoak21:
> The whole concept of major/minor can be hidden from callers by using a new `Version` class and only letting its constructor be aware of these, and let the functions work exclusively with this class, but that might complicate this PR too much at this point, so probably an overkill.


OK this sounds good! I dont think its so overkill here.

imap_data_access/file_validation.py
```diff
@@ -399,6 +411,18 @@ def extract_filename_components(filename: str | Path) -> dict:
 
         del components["interval_type"]
 
+        # Initialize version components
+        components["major_version"] = None
+        components["minor_version"] = None
+        version = components["version"]
```
lacoak21: Should we remove version entirely now? 

I think it may be nice to still have this here to return the string e.g. "vXXX" or "vMMM.mmmm"

tech3371: I would keep it still

imap_data_access/file_validation.py
```diff
@@ -45,6 +45,167 @@ def generate_imap_file_path(filename: str) -> ImapFilePath:
     )
 
 
+class Version:
```
lacoak21, line 50: @vineetbansal can you take a look at this class? Is there anything else its missing?

vineetbansal:
Thanks @lacoak21. Let me do a deep-dive into this PR and get back to you by tomorrow morning.

vineetbansal:
This is cool! I'm wondering what you think about the following changes, which are incremental over what you've already done here:

Right now `Version` is stateless. If it stored major/minor versions as instance variables then all callers could use it super cleanly, and one could do:
```
legacy_version = Version(None, 42)
assert legacy_version.major is None
assert legacy_version.minor == 42

version2 = Version(2, 1)
assert str(version1) == "v002.0001"

version3 = Version.from_version("v003.0002")

assert legacy_version < version2 < version3 

assert Version.from_version("xyz", raise_error=False) is None
```

So `.to_version()` can just be `__str__`, and `from_version` can remain a `staticmethod` that returns a `Version` object. A `__lt__` and `__eq__` along with [functools.total_ordering](https://docs.python.org/3/library/functools.html#functools.total_ordering) decorator will allow comparison/sorting of versions.

Everything else like `is_valid_version`, `is_valid_version_minor_only`, `_validate_range`, `version_regex` can remain static.

Sorry to keep suggesting more changes, but I think this will make it super useful for callers. If you're understandably busy with other stuff but generally agree that this is useful, we can merge this (its usable already) and I can send another PR tomorrow on top of this (since then my work with `imap-processing` will be near nil anyway :))

lacoak21:
> Sorry to keep suggesting more changes, but I think this will make it super useful for callers. If you're understandably busy with other stuff but generally agree that this is useful, we can merge this (its usable already) and I can send another PR tomorrow on top of this (since then my work with `imap-processing` will be near nil anyway :))

Ah I see. I was wondering about this! I had assumed from your earlier message you wanted it stateless but I can make these updates!

vineetbansal:
This is great and looks very clean. Thanks!

I think the last thing to think about is that Jon mentioned in a meeting (and he has also outlined this in his detailed google doc), that existing files should be interpreted as having version 1, or "v001." (He mentions "SPDF requires start at 1").

I don't know what the plan was to regenerate version 1 files - i.e. does the version [reported](https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/3266#issuecomment-4723390165) by `sds-data-manager` unconditionally return a major version 1 for now, or if they plan to use this API somehow. If its the former (which I think is the plan?), then everything here looks good. If the latter, then a helper method to go from legacy to new format might be needed here that they can use for the migration.

vineetbansal:
@lacoak21 - I noticed that `ReleaseFilePath` inherits from `AncillaryFilePath` whose `extract_filename_components` uses `rf"_(?P<version>{Version.minor_only_version_pattern})"` for the `version` component. Does something need to change for `ReleaseFilePath` ? I understand that `AncillaryFilePath`s behavior is correct since they will never get major versions, but I'm not sure if/where `ReleaseFilePath` objects are generated.

Also, `version_regex` and `is_valid_version` are referring to `valid_imap_version_pattern` instead of `VALID_VERSION_PATTERN`. If the intention was to have a single line to change in the future, maybe they should point to `VALID_VERSION_PATTERN`?

lacoak21:
> I don't know what the plan was to regenerate legacy files - i.e. does the version [reported](https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/3266#issuecomment-4723390165) by `sds-data-manager` unconditionally return a major version 1 for now, or if they plan to use this API somehow. If its the former (which I think is the plan?), then everything here looks good. If the latter, then a helper method to go from legacy to new format might be needed here that they can use for the migration.

I think we are good here? @tech3371 please let me know if that is not true in this case. Do we have a definite plan for this?

lacoak21:
> I think we are good here? @tech3371 please let me know if that is not true in this case. Do we have a definite plan for this?

I talked to tenzin and nothing needs to change here.

lacoak21:
> @lacoak21 - I noticed that `ReleaseFilePath` inherits from `AncillaryFilePath` whose `extract_filename_components` uses `rf"_(?P<version>{Version.minor_only_version_pattern})"` for the `version` component. Does something need to change for `ReleaseFilePath` ? I understand that `AncillaryFilePath`s behavior is correct since they will never get major versions, but I'm not sure if/where `ReleaseFilePath` objects are generated.
> 
> Also, `version_regex` and `is_valid_version` are referring to `valid_imap_version_pattern` instead of `VALID_VERSION_PATTERN`. If the intention was to have a single line to change in the future, maybe they should point to `VALID_VERSION_PATTERN`?

1. ReleaseFilePath should never have a major version and should inherit Version.minor_only_version_pattern
2. VALID_VERSION_PATTERN is part of the ScienceFilePath class and is more specific than the Version class. I think the version class should have the more general version pattern (including vXXX and vMMM.mmmm) and then the scienceFilePath can make it more specific. Did I understand that question correctly?

PR #326, opened 2026-06-10
Add global release table subparser 

lacoak21:
# Change Summary

## Overview
Add a sub parser command for querying the global release table for the latest release number.

### DIFF
--- a/imap_data_access/cli.py
+++ b/imap_data_access/cli.py
@@ -29,7 +29,7 @@ from imap_data_access.file_validation import (
     ScienceFilePath,
     generate_imap_file_path,
 )
-from imap_data_access.io import query, release, spice_query
+from imap_data_access.io import query, query_release_versions, release, spice_query
 from imap_data_access.utils import ReleaseType
 from imap_data_access.webpoda import download_daily_data
 
@@ -348,7 +348,12 @@ def _release_parser(args: argparse.Namespace):
     args : argparse.Namespace
         An object containing the parsed arguments and their values
     """
-    # All validation is now handled in io.py
+    if not getattr(args, "release_type", None):
+        raise ValueError(
+            "--release-type is required when submitting a release. "
+            "Run 'release -h' for more information."
+        )
+    # All other validation is handled in io.py
     release(
         instrument=args.instrument,
         release_type=args.release_type,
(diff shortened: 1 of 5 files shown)


imap_data_access/cli.py
```diff
@@ -696,7 +706,31 @@ def main():
         "release",
         help=release_help,
         formatter_class=argparse.RawTextHelpFormatter,
+        description=(
+            "Submit a release to the IMAP SDC, or use the 'query' subcommand\n"
+            "to look up the latest global release number.\n\n"
+            "Examples:\n"
+            "  imap-data-access release --release-type release ...\n"
+            "  imap-data-access release query\n"
```
lacoak21: Should this be 
```suggestion
            "  imap-data-access release latest\n"
```

tech3371:
```suggestion
            "  imap-data-access release --release-number latest\n"
```
or

```suggestion
            "  imap-data-access query --release-number latest\n"
```

lacoak21: Hm i think i will do  `imap-data-access release latest`. I think adding the --release-number flag implies that you can pass in a release number OR latest. Right now the only thing anyone would ever need is the latest and the api endpoint only finds the latest. If we end up needing to query by --release-number then we can add back that flag.

lacoak21: I just wanted to keep this as simple as possible without any user inputs.

tech3371: If I didn't have the context, `imap-data-access release latest` could mean I am releasing latest. I don't feel strongly about my suggestion either because it can be confusing too.

These are areas where I had trouble myself in the past couple weeks when coming up with release command and Tim has been huge help. I would ask him as second opinion. I recommend second opinion since this could stick for long time.

lacoak21: Ah that is a good point. I agree that could be confusing. Ok Ill send a message to infrastructure.

lacoak21: Ok see slack thread. 

We decided no changes were necessary and we should stick with what is here. 

https://lasp.slack.com/archives/C0B3Z7JF1QR/p1781192693528659

tech3371 (approved):

PR #328, opened 2026-06-16
Fix for failing tests

### LINKED ISSUE #327: Some tests do not pass
In a brand new environment where `imap-data-access` is installed, some tests fail because they assume that environment variables like `IMAP_API_KEY` and `IMAP_DATA_ACCESS_URL` etc. are not set (which they are for me, as they will be for most developers).

In my case:
```
FAILED tests/test_config_options.py::test_configuration_updates[DATA_DIR-default0-/test/path] - AssertionError: assert '/media/vineetb/T7' == '/media/vinee...a-access/data'
FAILED tests/test_config_options.py::test_configuration_updates[API_KEY-None-test-api-key] - AssertionError: assert 'be24...' == 'None'
```

I think the correct fix is to omit them when calling `subprocess.run` (which inherits all envvars) before asserting that their default value is what the test expects them to be.

vineetbansal:
Closes #327

The tests are assuming that they will always be run in an environment where the tested `IMAP_*` environment variables are unset. This might be true for the CI, but not true for the general development environment used for testing.

This fix omits any `IMAP_*` that are set in the host environment from being used in `subprocess.run`.

### DIFF
--- a/tests/test_config_options.py
+++ b/tests/test_config_options.py
@@ -31,8 +31,14 @@ def test_configuration_updates(config_var, default, expected):
         f"import imap_data_access; print(imap_data_access.config['{config_var}'])",
     ]
     # Default case
+    # Strip any IMAP_ overrides from the current environment so the built-in
+    # default is what gets used in `subprocess.run`.
+    default_env = {
+        key: value for key, value in os.environ.items() if f"IMAP_{config_var}" != key
+    }
     proc = subprocess.run(
         command,
+        env=default_env,
         capture_output=True,
         check=True,
         text=True,


tmplummer (commented):
This fix should work, but I am wondering if there should be a `session` scoped fixture that monkeypatches any existing IMAP environment variables. I am not familiar with the full test suite in this package, but it seems like existing env vars could have side effects in places other than just this test.

vineetbansal:
I feel that in this case adding a fixture will not be helpful - this is the only test that is testing to see if config variables change as a result of modifying envvars (plus it checks their original values - more on that below). So adding a fixture far away from the test which is only ever used here will obscure things. All other tests pass in the presence/absence of these envvars anyway. (the existing autouse fixtures and the tests rely on config values, not envvars directly).

What we can do is to remove the `default` values from the matrix, since the test should only care about whether config values change when they should, not what their default values were.

tmplummer (approved):
This is fine

PR #332, opened 2026-06-23
Remove deprecated vXXX handling for scienceFiles

lacoak21:
# Change Summary

## Overview
Remove all handling of deprecated version format vXXX for science files. Originally I had made this backwards compatible but since we will be renaming files during a downtime transition period we are able to directly move to the vMMM.mmm format at once.

### DIFF
--- a/imap_data_access/file_validation.py
+++ b/imap_data_access/file_validation.py
@@ -387,7 +387,8 @@ class ScienceFilePath(ImapFilePath):
         data_level: str,
         descriptor: str,
         start_time: str,
-        version: str,
+        major_version: int | None,
+        minor_version: int,
         extension: str = "cdf",
         repointing: int | str | None = None,
         cr: int | None = None,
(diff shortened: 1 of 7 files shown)


imap_data_access/file_validation.py
```diff
@@ -387,7 +387,8 @@ def generate_from_inputs(
         data_level: str,
         descriptor: str,
         start_time: str,
-        version: str,
```
lacoak21, line 390: Do you have an opinion on keeping this truly backwards compatible and keeping the option to pass in the version string? I am on the fence.

tech3371: option to pass in version

lacoak21: OK cool thanks. And your reasoning is to keep it backwards compatible?

tech3371: change of mind. Let's remove `vxxx` support to reduce confusion in the future. Since we will make official release after we complete testing new major and minor version with pre-release. That is enough.

Before, I was concern with someone install latest version of imap-data-access and not seeing `vxxx` or `vMMM.mmmm`. Since we are using pre-release to test new version and have downtime to make that transition and if we announce to team well in advance, let's not make version to be backward compatible.

tech3371 (approved):
LGTM

